PR #29, opened 2022-05-04
[checkpoint] Copy over consolidate script

stephenroller:
**Patch Description**
Copy over the consolidation script for HF.

**Testing steps**
Used the script internally yesterday.


suchenzang (approved):
Add something for this in the docs somewhere plz

PR #30, opened 2022-05-04
BF16 support

ngoyal2707:
**Patch Description**
Adds support for bfloat16 via the --bf16 flag.

**Testing steps**
New CI. Manual testing.


metaseq/models/distributed_model.py
```diff
@@ -63,7 +63,10 @@ def DistributedModel(args, model, process_group, device):
         assert isinstance(model, FSDP), "expected model to already be wrapped in FSDP"
         wrapped_model = model
         if args.memory_efficient_fp16:
-            wrapped_model = wrapped_model.half()
+            if args.bf16:
```
suchenzang, line 66: Nit: This seems potentially confusing to future users. Seem like we need both `memory_efficient_fp16` _and_ `bf16` turned on here, to enable the following bf16 logic.  Should we rename the `memory_efficient_fp16` flag to something more descriptive?

metaseq/models/transformer.py
```diff
@@ -383,6 +383,14 @@ def __init__(self, args, dictionary, embed_tokens, no_encoder_attn=False):
         initialize_params_on_gpu = getattr(
             args, "tensor_parallel_init_model_on_gpu", False
         )
+        def floating_point_precision_convertor(x):
```
suchenzang, line 386: Ditto from previous nit, re: confusing flag setting of both `memory_efficient_fp16` and `bf16`.

metaseq/modules/transformer_layer.py
```diff
@@ -261,6 +262,12 @@ def __init__(
         )
         self.normalize_before = args.decoder_normalize_before
 
+        def floating_point_precision_convertor(x):
```
suchenzang, line 265: Nit: move this to a utils file somewhere to not have two copies of the same function defined in two places.

metaseq/models/distributed_model.py
stephenroller: +1 let's move this outside the args.mem_eff_fp16 gate

suchenzang:
Woo! Thanks for porting this out to the public repo! Just some nits about naming and staying DRY (and getting the tests to pass), but otherwise looks good!

metaseq/models/distributed_model.py
suchenzang: @ngoyal2707 let's solve the backwards compatibility breakage separately - we need a lot more love to the checkpointing logic, so fixing the "breakage" would be a good starting point too!

metaseq/trainer.py
```diff
@@ -1035,6 +1035,8 @@ def _prepare_sample(self, sample, is_dummy=False):
 
         def apply_half(t):
```
m3rlin45, line 1036: This function seems surprising now, the return value is only partly config driven

| config value | Input Value | Return Value | |
| ------ | ----- | -----| --- |
| bf16 | fp32 | bf16 | inconsistent with function name |
| bf16 | half | half | **inconsistent with config** |


I think the naming issue is less concerning, (naming is hard) but in the case where the config is bf16, and the input value is half, I think it is less surprising if we return bf16.

metaseq/models/distributed_model.py
stephenroller: I've just added an assert into MemoryEfficientFP16Optimizer that asserts bf16 isn't true and tells the user not to mix the two.

metaseq/trainer.py
stephenroller: does this seem better now?

metaseq/trainer.py
```diff
@@ -1033,13 +1033,16 @@ def _prepare_sample(self, sample, is_dummy=False):
         if self.cuda:
             sample = utils.move_to_cuda(sample)
 
-        def apply_half(t):
-            if t.dtype is torch.float32:
+        def apply_precision_conversion(t):
```
suchenzang, line 1036: The end result of this function and `floating_point_precision_convertor` seem to be the same. What's the reason for two methods here? 

Also it seems like goal is to always halve / lower the precision. Another naming nit: add "lower" or "half" to the name (going back to apply_half lol)

metaseq/models/distributed_model.py
suchenzang: Cool, that works

metaseq/trainer.py
m3rlin45: naming is hard. I would bias towards something like `lower_precision` or `apply_16bit`

I'm happy with the logic changes (we don't ignore the config in the case of fp16 in).

suchenzang:
Looks like commits [bce9622](https://github.com/facebookresearch/metaseq/pull/30/commits/bce9622de64b1eab9b9705a1a5a32f7ce512dc71)..[eb96da5](https://github.com/facebookresearch/metaseq/pull/30/commits/eb96da5462b76b46f751b862e2ba79eb3d4a9247) might've broken something. 

Flagged the below assertion error when benchmarking:
```
AssertionError: expects all parameters to have same dtype: fp32: _fpw_module.self_attn.qkv_proj,_fpw_module.self_attn.qkv_proj,_fpw_module.self_attn.out_proj,_fpw_module.self_attn.out_proj,_fpw_module.self_attn_layer_norm,_fpw_module.self_attn_layer_norm,_fpw_module.fc1,_fpw_module.fc1,_fpw_module.fc2,_fpw_module.fc2,_fpw_module.final_layer_norm,_fpw_module.final_layer_norm
```

stephenroller:
Works now, tested with `python metaseq/launcher/opt_baselines.py --local -p foo --model-size 125m -g 8 --azure`.

metaseq/utils.py
```diff
@@ -607,6 +615,29 @@ def get_random_port():
...
+):
+    """
+    Convert a tensor x into the desired dtype.
+
+    Also sanity checks combinations of options.
+    """
+    if memory_efficient_fp16:
+        assert not bf16, "Do not combined bf16 with memory_efficient_fp16."
+    if bf16:
+        assert fp16, "Setting --bf16 requires also setting --fp16 for legacy reasons."
```
suchenzang, line 629: ;_;

suchenzang (approved):

metaseq/utils.py
stephenroller: hey, at least it doesn't just silently do the wrong thing :P

metaseq/models/distributed_model.py
emilydinan: I actually am confused here lol -- why is this under `if args.memory_efficient_fp16`? cc @ngoyal2707 @suchenzang @stephenroller

PR #38, opened 2022-05-04
changed metaseq_internal to metaseq

tsor13:
**Patch Description**
In the shard script, it incorrectly pointed to `metaseq_internal` instead of `metaseq` as the location of the `reshard_mp` file.

**Testing steps**
I tested my change by sharding the 175b file with the modified script.


suchenzang (approved):
Thank you!

PR #48, opened 2022-05-05
CLI interactive

stephenroller:
**Patch Description**
Adds support for CLI interactive.

This involves a major change to distributed initialization, which I copied from ParlAI: Instead of having the main process spawn N subprocesses all running the distributed code, and then idle until they wrap up, we have it spawn N-1 processes and then we also run the distributed code on the main process.

This means that exceptions or signals like KeyboardInterrupts in the main thread can be caught and propagated to the spawns to be terminated. It also means we don't need to do any hacks to reroute stdin/stdout in the rank0 process.

**Testing steps**
Screen Shot 2022-05-05 at 16 11 46

Also tested that I can cleanly ctrl-c now, and also several exceptions due to standard iterations were correctly propagated.

### DIFF
--- a/metaseq/distributed/utils.py
+++ b/metaseq/distributed/utils.py
@@ -238,7 +238,7 @@ def _spawn_helper(main, cfg, kwargs):
         for p in spawncontext.processes:
             if p.is_alive():
                 os.kill(p.pid, signal.SIGTERM)
-            raise
+        raise
 
 
 def call_main(cfg: MetaseqConfig, main, **kwargs):
--- a/metaseq/distributed/utils.py
+++ b/metaseq/distributed/utils.py
@@ -181,7 +181,8 @@ def distributed_init(cfg: MetaseqConfig):
 
 def distributed_main(i, main, cfg: MetaseqConfig, kwargs):
     if not cfg.distributed_training.distributed_no_spawn:
-        # local spawner is actually rank + 1
+        # if in local spawning, i is offset by -1 since torch.multiprocessing.spawn
+        # always starts at rank 0
         i = i + 1
     cfg.distributed_training.device_id = i
     if torch.cuda.is_available() and not cfg.common.cpu:
@@ -191,7 +192,7 @@ def distributed_main(i, main, cfg: MetaseqConfig, kwargs):
         # able to pass local rank.
         os.environ["LOCAL_RANK"] = str(cfg.distributed_training.device_id)
     if cfg.distributed_training.distributed_rank is None:
-        # _span_helper
+        # start_rank is the rank of gpu 0 on this machine.
         cfg.distributed_training.distributed_rank = kwargs.pop("start_rank", 0) + i
 
     cfg.distributed_training.distributed_rank = distributed_init(cfg)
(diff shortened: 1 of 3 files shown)


suchenzang:
Does your test include completion of that test prompt?

stephenroller:
> Does your test include completion of that test prompt?

Yes it generates lol. But it generates a lot.

metaseq/distributed/utils.py
```diff
@@ -179,14 +180,18 @@ def distributed_init(cfg: MetaseqConfig):
...
     cfg.distributed_training.device_id = i
     if torch.cuda.is_available() and not cfg.common.cpu:
         torch.cuda.set_device(cfg.distributed_training.device_id)
         # This is temporary way of making microsoft Tutel happy, as it reads the local rank from
         # the env. To make it work in cleaner way, we might need to change their interfaces to be
         # able to pass local rank.
         os.environ["LOCAL_RANK"] = str(cfg.distributed_training.device_id)
-    if cfg.distributed_training.distributed_rank is None:  # torch.multiprocessing.spawn
+    if cfg.distributed_training.distributed_rank is None:
+        # _span_helper
```
suchenzang, line 194: What is this?

metaseq/distributed/utils.py
```diff
@@ -179,14 +180,18 @@ def distributed_init(cfg: MetaseqConfig):
 
 
 def distributed_main(i, main, cfg: MetaseqConfig, kwargs):
+    if not cfg.distributed_training.distributed_no_spawn:
```
suchenzang, line 183: Can we remove `distributed_no_spawn` config entirely? Will we have `ntasks_per_node` > 1?

metaseq/distributed/utils.py
```diff
@@ -200,6 +205,43 @@ def distributed_main(i, main, cfg: MetaseqConfig, kwargs):
...
+            cfg.distributed_training.distributed_world_size,
+        ),
+        join=False,
+        start_method='spawn',
+    )
+
+    try:
+        # -1 because we offset by +1 inside distributed_main when using
+        # spawn_helper
+        retval = distributed_main(-1, main, cfg, kwargs)
```
suchenzang, line 232: Main motivation behind getting rid of `distributed_no_spawn` gate is to simplify things like this.

metaseq_cli/interactive_cli.py
```diff
@@ -0,0 +1,121 @@
...
+    logger.debug(f"Input: {inp}")
+    return "\n".join(inp)
+
+
+def worker_main(cfg1: MetaseqConfig, namespace_args=None):
+    global generator
+    # make sure generations are stochastic since we have many workers
+    torch.manual_seed(random.randint(1, 20000))
+    torch.cuda.manual_seed(random.randint(1, 20000))
+    cfg = cfg1
```
suchenzang, line 67: Why do you have to do this copy?

metaseq/distributed/utils.py
stephenroller: No, DNS is great and useful.

metaseq/distributed/utils.py
stephenroller: This would still be the case. The 0-indexing comes from torch.multiprocessing.spawn.

metaseq/distributed/utils.py
stephenroller: DNS is particularly useful during our training runs or our API runs, as it let's any exceptions in the subworkers get noticed by slurm and kills (or requeues) the entire job. If you use spawning, then there's a good chance an exception anywhere OUTSIDE of the local rank 0 will result in a hang.

suchenzang: When is DNS == False useful then?

stephenroller: Local command line, because now you don't get the luxury of SLURM handling launching.

suchenzang (approved):

Issue #14, opened 2022-03-29
Add MyPy type checking to CI


m3rlin45:
## 🚀 Feature Request
let's have typechecking, like in ParlAI: https://github.com/fairinternal/ParlAI-Internal/blob/705258b5799e69ef8942a8fc56928ad779e0980b/.github/workflows/lint.yml

### Motivation
It would be good to have some type checks (as suggestions, not a blocking test) to help catch common errors

### Pitch

probably can just copy/past from ParlAI:
https://github.com/fairinternal/ParlAI-Internal/blob/705258b5799e69ef8942a8fc56928ad779e0980b/.github/workflows/lint.yml

https://fb.workplace.com/100020258253907/posts/744791912872744/
Some slides about the ParlAI version

TommyDew42:
Hi! It looks like a very good issue I can contribute. Is it possible I can pick this up?

stephenroller:
Please Tommy, we would welcome!

punitkoura:
Looks like this is done, closing

PR #103, opened 2022-05-17
Add MyPy to CI

TommyDew42:
**Patch Description**
PR for [issue 14](https://github.com/facebookresearch/metaseq/issues/14).

Here we add MyPy to CI.

**Testing steps**
Run `mypy` at local laptop & run the check via an actual pull request.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,3 +44,8 @@ jobs:
       run: |
         pip install black
         black --check .
+
+    - name: Type Checking with MyPy
+      run:  |
+        pip install mypy
+        mypy
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -48,4 +48,4 @@ jobs:
     - name: Type Checking with MyPy
       run:  |
         pip install mypy
-        mypy
+        mypy || true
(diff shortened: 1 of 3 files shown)


TommyDew42:
@stephenroller 

I run the CI in the [PR](https://github.com/TommyDew42/metaseq/pull/1) in my fork repo.
But there are a lot of `mypy` type checking [errors](https://github.com/TommyDew42/metaseq/runs/6474179539?check_suite_focus=true).

How do we do about them?

.github/workflows/lint.yml
```diff
@@ -44,3 +44,8 @@ jobs:
       run: |
         pip install black
         black --check .
+
+    - name: Type Checking with MyPy
+      run:  |
+        pip install mypy
+        mypy
```
stephenroller, line 51: Lets make this || true for now. Mypy is best as a warning not an error in its current state

mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+python_version = 3.7
```
suchenzang, line 2: I believe we have python 3.8 dependency in our CI setup - could we change this to 3.8 instead?

suchenzang (approved):

stephenroller:
Thanks for the contribution.

PR #104, opened 2022-05-17
Support document level attention while training model parallel models

punitkoura:
**Patch Description**
We are trying to optimize model training by having document level attention. What this entails is that we modify the attention mask to just attend to tokens in the current document, as opposed to having a constant triangular (causal) attention mask. The goal here is to ensure that tokens don't attend across documents.

To accomplish this, we
1. Change the constant upper triangular mask to depend on the data separator tokens, specified by the `self_attn_doc_sep` field
2. Reset the positional encoding at the end of each document.

**Testing steps**
Consider a list of 20 documents. We consider the following two cases - 
1. Passing the documents, one per sequence, in a batch of 20 sequences. No special attention mask is used, we still use the upper triangular mask.
2. Passing the documents packed in a single sequence, separated by a special document separator token. Document attention is used here.

By intuition, we can see that the gradients in these two scenarios should match. We find that it is indeed the case.

Documents packed, with document level attention
```
> /fsx-mudslide/punitkoura/src/metaseq/metaseq/trainer.py(781)train_step()
-> grad_norm = self.clip_grad_norm(
(Pdb) self.model.decoder.embed_positions.weight.grad[2:10]
tensor([[-6.5479e-03,  9.6962e-03,  2.6375e-03,  ..., -5.1121e-03,
         -1.9985e-02,  2.4901e-03],
        [ 2.5498e-02, -7.1761e-03,  2.2065e-02,  ...,  6.4638e-03,
         -1.0154e-02,  5.8792e-03],
        [ 7.9259e-05, -4.0577e-03,  3.9698e-03,  ...,  2.3798e-04,
          3.1224e-03,  1.0833e-02],
        ...,
        [-7.4849e-06,  1.7103e-03, -1.5433e-03,  ...,  2.2430e-03,
         -9.3732e-03,  1.5123e-03],
        [-1.4408e-03, -1.8149e-03,  6.9925e-06,  ...,  1.1794e-03,
         -1.4160e-03,  2.6913e-03],
        [-1.4234e-03, -6.3481e-03,  7.6257e-03,  ...,  7.9472e-03,
         -1.1467e-03,  3.3719e-03]], device='cuda:0')
(Pdb) 
```

Documents sent one by one
```
> /fsx-mudslide/punitkoura/src/metaseq/metaseq/trainer.py(781)train_step()
-> grad_norm = self.clip_grad_norm(
(Pdb) self.model.decoder.embed_positions.weight.grad[2:10]
tensor([[-6.5555e-03,  9.6926e-03,  2.6306e-03,  ..., -5.1131e-03,
         -1.9988e-02,  2.4959e-03],
        [ 2.5496e-02, -7.1761e-03,  2.2073e-02,  ...,  6.4701e-03,
         -1.0157e-02,  5.8840e-03],
        [ 7.5817e-05, -4.0496e-03,  3.9688e-03,  ...,  2.3734e-04,
          3.1189e-03,  1.0830e-02],
        ...,
        [-1.3878e-05,  1.7050e-03, -1.5432e-03,  ...,  2.2473e-03,
         -9.3719e-03,  1.5083e-03],
        [-1.4369e-03, -1.8114e-03,  5.8990e-06,  ...,  1.1820e-03,
         -1.4172e-03,  2.6962e-03],
        [-1.4229e-03, -6.3467e-03,  7.6208e-03,  ...,  7.9439e-03,
         -1.1527e-03,  3.3779e-03]], device='cuda:0')
(Pdb) 
```

### DIFF
--- a/metaseq/model_parallel/modules/multihead_attention.py
+++ b/metaseq/model_parallel/modules/multihead_attention.py
@@ -22,7 +22,7 @@ try:
         RowParallelLinear,
         split_tensor_along_last_dim,
     )
-    from megatron.model.fused_softmax import ScaledUpperTriangMaskedSoftmax
+    from megatron.model.fused_softmax import ScaledUpperTriangMaskedSoftmax, ScaledMaskedSoftmax
     from megatron.model import utils as megatron_utils
 
     has_megatron_submodule = True
@@ -373,10 +373,21 @@ class ModelParallelMultiheadAttention(nn.Module):
             if attn_mask is not None:
                 matmul_result = torch.nan_to_num(matmul_result)
 
-            # attention_scores = matmul_result.view(*output_size)
-            attn_probs = ScaledUpperTriangMaskedSoftmax.apply(matmul_result, 1.0)
-            # attn_probs = self.scale_mask_softmax(attention_scores,
-            #                                         attn_mask)
+            if len(attn_mask.size()) == 3:
+                # Going back to original scaled_masked_softmax to accomodate
+                # non-causal attention masking (use the given input attention)
+                attention_scores = matmul_result.view(*output_size)
+                attn_mask = attn_mask < -0.5
+                attn_mask = attn_mask.unsqueeze(1)
+                attn_probs = ScaledMaskedSoftmax.apply(attention_scores, attn_mask, 1.0)
+                attn_probs = attn_probs.view(output_size[0]*output_size[1],
+                    output_size[2],
+                    output_size[3])
+            else:
+                # attention_scores = matmul_result.view(*output_size)
+                attn_probs = ScaledUpperTriangMaskedSoftmax.apply(matmul_result, 1.0)
+                # attn_probs = self.scale_mask_softmax(attention_scores,
+                #                                         attn_mask)
             with get_cuda_rng_tracker().fork():
                 attn_probs = self.dropout_module(attn_probs)
 
--- a/metaseq/model_parallel/modules/multihead_attention.py
+++ b/metaseq/model_parallel/modules/multihead_attention.py
@@ -22,7 +22,10 @@ try:
         RowParallelLinear,
         split_tensor_along_last_dim,
     )
-    from megatron.model.fused_softmax import ScaledUpperTriangMaskedSoftmax, ScaledMaskedSoftmax
+    from megatron.model.fused_softmax import (
+        ScaledUpperTriangMaskedSoftmax,
+        ScaledMaskedSoftmax,
+    )
     from megatron.model import utils as megatron_utils
 
     has_megatron_submodule = True
@@ -373,6 +376,11 @@ class ModelParallelMultiheadAttention(nn.Module):
             if attn_mask is not None:
                 matmul_result = torch.nan_to_num(matmul_result)
 
+            # The attn_mask shape can be either seq_length x seq_length, or batch_size x seq_length x seq_length
+            # depending on whether we are broadcasting the same attention mask across the batch, or
+            # masking dynamically based on the data (e.g. for document attention).
+            # If we have a per sequence mask, the condition len(attn_mask.size()) == 3
+            # is true.
             if len(attn_mask.size()) == 3:
                 # Going back to original scaled_masked_softmax to accomodate
                 # non-causal attention masking (use the given input attention)
@@ -380,9 +388,9 @@ class ModelParallelMultiheadAttention(nn.Module):
                 attn_mask = attn_mask < -0.5
                 attn_mask = attn_mask.unsqueeze(1)
                 attn_probs = ScaledMaskedSoftmax.apply(attention_scores, attn_mask, 1.0)
-                attn_probs = attn_probs.view(output_size[0]*output_size[1],
-                    output_size[2],
-                    output_size[3])
+                attn_probs = attn_probs.view(
+                    output_size[0] * output_size[1], output_size[2], output_size[3]
+                )
             else:
                 # attention_scores = matmul_result.view(*output_size)
                 attn_probs = ScaledUpperTriangMaskedSoftmax.apply(matmul_result, 1.0)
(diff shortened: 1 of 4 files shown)


metaseq/models/transformer.py
```diff
@@ -561,10 +562,35 @@ def forward_embedding(
         incremental_state: Optional[Dict[str, Dict[str, Optional[Tensor]]]] = None,
     ):
         # embed tokens and positions
-        positions = None
+        if self.self_attn_doc_sep != -1:
+            # create own positions when self_attn_doc_sep is set
```
stephenroller, line 566: I dont understand why this differs from the normal case?

punitkoura: We are essentially resetting positions based on document separator tokens...

For instance, if the doc separator is 2, and the tokens are 
143 345 2 5435 2
The default positions would be
0 1 2 3 4

But with document level attention, we would like to reset positions as well on sentence boundaries. So, the positions become
0 1 2 0 1

metaseq/models/transformer.py
```diff
@@ -561,10 +562,35 @@ def forward_embedding(
...
+                    if k == 0:
+                        mask_with_reset[batch_idx, doc_sep_idx] = -doc_sep_idx + 1
+                    else:
+                        mask_with_reset[batch_idx, doc_sep_idx] = (
+                            batch_doc_indices[k - 1] - doc_sep_idx + 1
+                        )
+            positions = (
+                torch.cumsum(mask_with_reset, dim=1).type_as(mask) * mask
+            ).long() + self.padding_idx
+            # HACK set padding_idx to None to work
```
suchenzang, line 586: A bit longer comment for why this hack is needed?

metaseq/models/transformer.py
```diff
@@ -561,10 +562,35 @@ def forward_embedding(
         incremental_state: Optional[Dict[str, Dict[str, Optional[Tensor]]]] = None,
     ):
         # embed tokens and positions
-        positions = None
+        if self.self_attn_doc_sep != -1:
```
suchenzang, line 565: Can we move out this hard-coded sentinel value and assign it a variable name? Will be cleaner to refactor later on if that ever comes up.

metaseq/models/transformer.py
```diff
@@ -561,10 +562,35 @@ def forward_embedding(
...
+            # create own positions when self_attn_doc_sep is set
+            mask = tokens.ne(self.padding_idx).int()
+            mask_with_reset = tokens.ne(self.padding_idx).int()
+            mask_with_reset[:, :] = 1
+            doc_id_indices = (tokens == self.self_attn_doc_sep).nonzero().tolist()
+            for batch_idx in range(tokens.size(0)):
+                batch_doc_indices = [
```
suchenzang, line 572: Nit: add a comment for what this logic is doing (separating intention vs the implementation) to make it easier to parse what nested for loops are trying to loop over.

metaseq/models/transformer.py
stephenroller: Hm. But if we're going this far, why not just have properly separate batch examples? The padding overhead, if implemented with heuristics, doesn't remotely approach high enough to worry, no?

ramakanth-pasunuru: When we pack multiple examples in a sequence, we can efficiently use the full sequence length (2048), thereby each batch can accommodate more examples in comparison to having separate batch examples?

metaseq/models/transformer.py
punitkoura: Added more context.

metaseq/model_parallel/modules/multihead_attention.py
```diff
@@ -367,10 +370,26 @@ def forward(
...
+                # non-causal attention masking (use the given input attention)
+                attention_scores = matmul_result.view(*output_size)
+                attn_mask = attn_mask < -0.5
+                attn_mask = attn_mask.unsqueeze(1)
+                attn_probs = ScaledMaskedSoftmax.apply(attention_scores, attn_mask, 1.0)
+                attn_probs = attn_probs.view(
+                    output_size[0] * output_size[1], output_size[2], output_size[3]
+                )
+            else:
+                # attention_scores = matmul_result.view(*output_size)
```
suchenzang, line 389: Clean up commented out code?

metaseq/models/transformer.py
```diff
@@ -797,13 +846,29 @@ def buffered_future_mask(self, tensor):
             if self.use_alibi:
                 alibi = self.alibi.repeat(batch_size, 1, 1)  # batch_size, 1, 1
                 self._future_mask = self._future_mask.unsqueeze(0) + alibi
+            elif self.self_attn_doc_sep != UNSPECIFIED_DOC_SEP:
```
suchenzang, line 849: Should this conditional be based on not using alibi? Or be standalone?

metaseq/models/transformer.py
```diff
@@ -561,10 +566,53 @@ def forward_embedding(
...
+                batch_doc_indices = [
+                    index[1] + 1
+                    for index in doc_id_indices
+                    # index[1] + 1 < tokens.size(1) to prevent overflow
+                    if index[0] == batch_idx and index[1] + 1 < tokens.size(1)
+                ]
+                batch_doc_indices.sort()
+                for k, doc_sep_idx in enumerate(batch_doc_indices):
+                    if k == 0:
+                        mask_with_reset[batch_idx, doc_sep_idx] = -doc_sep_idx + 1
```
suchenzang, line 598: All of the +1 and -1 logic scattered throughout this chunk of code is a bit difficult to parse / understand why. Can you add a comment on what these offsets are supposed to account for?

metaseq/model_parallel/modules/multihead_attention.py
punitkoura: Sorry! Removed the commented out code

metaseq/models/transformer.py
punitkoura: Good point, fixed this.

metaseq/models/transformer.py
punitkoura: Added a comment to explain the reason

suchenzang:
Generally LGTM! Can we get a quick baseline training run check to make sure loss curve still looks sane?

punitkoura:
@suchenzang The loss and activation norm curves for training with document attention seem okay


Screen Shot 2022-06-17 at 11 53 09 AM
Screen Shot 2022-06-17 at 11 52 57 AM

suchenzang (approved):

suchenzang:
Ah one last sanity check: what's the wps cost of doing document level attention in this current implementation (vs not)?

PR #107, opened 2022-05-19
Add eos mode for src tgt dataset

sriniiyer:
Description:
Currently, multiple examples are packed into a single sequence during training/fine-tuning and attention can happen across sequences. This diff adds eos mode, which uses a single example per sequence. This can be useful for fine-tuning/debugging. 

Test Plan:
Code path is executed via --break-mode eos_pad_8 and code has been tested earlier in a private branch.

### DIFF
--- a/metaseq/data/streaming_src_tgt_dataset.py
+++ b/metaseq/data/streaming_src_tgt_dataset.py
@@ -51,7 +51,7 @@ class StreamingSrcTgtDataset(torch.utils.data.IterableDataset):
         self.seed = seed
         if break_mode == "none" or break_mode == "complete":
             self.block_iterator = yield_src_tgt_blocks
-        elif break_mode == "eos_pad_8": # Single example per sequence
+        elif break_mode == "eos_pad_8":  # Single example per sequence
             self.block_iterator = yield_src_tgt_single_sentences_pad_8
         else:
             raise NotImplementedError(f"Unknown break mode: {break_mode}")
@@ -165,6 +165,7 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
             "src_block": src_block,
             "tgt_block": tgt_block,
         }
+
 
 def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, padding_idx):
     """Mimics sample-break-mode eos i.e. 1 example per sequence without any packing.
--- a/metaseq/data/streaming_src_tgt_dataset.py
+++ b/metaseq/data/streaming_src_tgt_dataset.py
@@ -7,6 +7,7 @@ from typing import Optional
 
 import numpy as np
 import torch
+import math
 
 
 class StreamingSrcTgtDataset(torch.utils.data.IterableDataset):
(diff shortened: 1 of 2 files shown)


metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -164,3 +165,45 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
             "src_block": src_block,
             "tgt_block": tgt_block,
         }
+
+
+def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, padding_idx):
```
suchenzang, line 170: Is there a specific reason why it has to be `8` in particular? Can this be generalized to any value (or will that never be needed)?

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -164,3 +165,45 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
...
+            tgt = tgt[:block_size]
+
+        cur_src_block.append(src)
+        cur_src_block_ids.append(idx)
+        cur_tgt_block.append(tgt)
+
+        # We round up to a multiple of 8 + 1, because later on
+        # one element is removed for src/target tensor creation
+        # which brings it back to a multiple of 8. block_size is
+        # already passed with + 1 included.
```
suchenzang, line 197: Add an assertion for `block_size` being passed with + 1 included?

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -415,7 +415,6 @@ def get_batch_iterator(
         # thus increasing randomness. This assumes that no single document spans
         # 10 full batches, which is reasonable when batch sizes are in the
         # millions and documents are on average much smaller.
-        assert isinstance(dataset, StreamingTokenBlockDataset)
```
suchenzang, line 418: This assertion should be kept.  You just need to include `StreamingSrcTgtDataset` instance, right?

metaseq/data/streaming_src_tgt_dataset.py
sriniiyer: 8 worked well across the most number of models and did not trigger the internal assert from megatron. Once document attention is debugged and landed, we can use that instead of this and this will just become a fallback mechanism in case document attention gives errors.

metaseq/data/streaming_src_tgt_dataset.py
sriniiyer: Hard to know that here, block size is passed in streaming_language_modeling task with the +1.

metaseq/tasks/streaming_language_modeling.py
sriniiyer: Good point. Will add this

metaseq/data/streaming_src_tgt_dataset.py
suchenzang: Is there a way to assert something around this?  This seems like it depends on very brittle logic...

sriniiyer: Technically, it will trigger the megatron assert - so it will not be something that will go unnoticed and cause issues. Not sure how to assert for this.

metaseq/tasks/streaming_language_modeling.py
sriniiyer: Done

suchenzang (approved):

PR #108, opened 2022-05-19
Allows download into a target directory

StellaAthena:
**Patch Description**
Minor modification to the 175B download script to allow the user to download the files into a directory other than the one that the command is being run from.

**Testing steps**
I ran it.

### DIFF
--- a/metaseq/scripts/download_opt175b.sh
+++ b/metaseq/scripts/download_opt175b.sh
@@ -17,7 +17,7 @@ do
   for shard_id in $(seq 0 123)
   do
     filename="checkpoint_last-model_part-$part_id-shard$shard_id.pt"
-    wget -O "$filename" "${presigned_url/$str_to_replace/$filename}"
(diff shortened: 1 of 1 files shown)


metaseq/scripts/download_opt175b.sh
```diff
@@ -17,7 +17,7 @@ do
   for shard_id in $(seq 0 123)
   do
     filename="checkpoint_last-model_part-$part_id-shard$shard_id.pt"
-    wget -O "$filename" "${presigned_url/$str_to_replace/$filename}"
+    wget -O "$2$filename" "${presigned_url/$str_to_replace/$filename}"
```
suchenzang, line 20: Nit: add a comment above indicating the use of this second arg?

StellaAthena: Ooops! I did add that comment to my repo but it seems to have happened in another branch. I’ll chose this PR, merge on my end, and reopen it

PR #109, opened 2022-05-20
Add Example Proportional Sampling Maximum

sriniiyer:
Add sampling from multiple corpora based on their size with a maximum size parameter as in Raffel et al., 2020, Wei et al., 2022, etc.

Test Plan
Tested with fine-tuning on various datasets with help from @qingliu1111

### DIFF
--- a/metaseq/tasks/streaming_language_modeling.py
+++ b/metaseq/tasks/streaming_language_modeling.py
@@ -83,6 +83,12 @@ class StreamingLanguageModelingConfig(MetaseqDataclass):
             "help": "smoothing alpha for sample rations across multiple datasets"
         },
     )
+    multicorpus_sampling_maximum: Optional[float] = field(
+        default=-1,
+        metadata={
+            "help": "Maximum size for example proportional sampling"
+        },
+    )
 
     # TODO common vars below add to parent
     seed: int = II("common.seed")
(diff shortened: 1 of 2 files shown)


metaseq/tasks/streaming_language_modeling.py
```diff
@@ -174,9 +178,15 @@ def _get_sample_prob(self, dataset_lens):
         """
         Get smoothed sampling porbability by corpus. This helps small corpus by upsampling them.
         """
-        prob = dataset_lens / dataset_lens.sum()
-        smoothed_prob = prob**self.args.multicorpus_sampling_alpha
-        smoothed_prob = smoothed_prob / smoothed_prob.sum()
+        if self.args.multicorpus_sampling_maximum == -1:
```
suchenzang, line 181: Nit: extract out the default value check to a variable, so the two can be coupled (aka define something like `DEFAULT_MULTICORPUS_MAX = -1` somewhere) ? Also what is a reasonable range to set the `multicorpus_sampling_maximum` value?

qingliu1111:
## Test Plan
Test this PR by reproducing results from the old branch `streamingft_new` that we used for FLAN fine-tuning. 
* 1.3B model
  * Old branch `streamingft_new`: acc 32.3605
  * This PR `sviyer/eps4`: acc 32.5812
* 175B
  *  Old branch `streamingft_new`:  acc 54.6536
  *  This PR `sviyer/eps4`:  
      * Same HP: acc 53.6358
      * Reduced lr: acc 54.4329

The change in this PR on the top of latest metaseq can reproduce experiment results got from old branch, and for large models small tweaks in HP is needed (might be due to the change in data order, which large models are more susceptible to ).

metaseq/tasks/streaming_language_modeling.py
sriniiyer: Done. The range would vary depending on the sizes of the datasets being sampled from. Literature has used values like 3000 (We et al., 2021), 65536 etc.

qingliu1111 (approved):

suchenzang (approved):

PR #113, opened 2022-05-20
Allow downloading to a target directory

StellaAthena:
**Patch Description**
Minor modification to the 175B download script to allow the user to download the files into a directory other than the one that the command is being run from.

**Testing steps**
I ran it.


stephenroller (approved):
Cheers

PR #116, opened 2022-05-22
Include diff with black check

suchenzang:


suchenzang:
Still don't know why my local black check didn't pick this up.

stephenroller (approved):

PR #120, opened 2022-05-23
Add Dockerfile

salanki:
**Patch Description**
Adds a Dockerfile to the project repository, based on the setup instructions to make it possible to run on any nvidia-docker enabled host and in Kubernetes.

Can be run locally with an interactive shell:
```
docker run --rm -it --gpus 1 coreweave/ml-images:metaseq-6cd6a708 bash
```

**Testing steps**
Build and run tested. Not yet tested for large inference or training.


suchenzang:
Seems reasonable enough - thanks for the PR!

suchenzang (approved):

PR #124, opened 2022-05-25
change omegaconf function name

toanngosy:
**Patch Description**
In omegaconf==2.2.1,  `_utils.is_primitive_type` changes to `_utils.is_primative_type_annotation`


suchenzang:
Thanks for the PR!  We will pin omegaconf to 2.1.1 instead for now (otherwise will be caught having to keep making these updates with any future breaking upgrades to omegaconf).

PR #126, opened 2022-05-27
Add missing Namespace import

todpole3:
Without this `gpt3_eval` hits the following exception:

```
NameError: name 'Namespace' is not defined
```


suchenzang:
Rerunning the failed GPU tests here, though I don't think that should be related to this diff.  @[chrislengerich](https://github.com/chrislengerich) I think you're hitting this error as well (we should really fix evals to not require doing this import here as a dependency, but leaving this here for now to unblock).

Issue #142, opened 2022-06-07
Requeue with restore file and broken checkpoint upload... is broken


suchenzang:
Right now, it seems like if we requeue a job that started with a restore file, and since starting from the restore file there has been broken checkpoint uploads, the run will simply restart from scratch while continuing to increment its iteration count.

Go through our checkpointing spagetti and figure out how to clean this up: https://github.com/facebookresearch/metaseq/blob/ae825b2fa9010ab0406f20d6164ebb058a7e97cf/metaseq/checkpoint_utils.py#L257

suchenzang:
So in the failure we saw, the following happened:
* `--restore-file` existed, and job trained from checkpoint 6500 until ~6600.
* During this time, no new checkpoints were uploaded, since save interval is set at 250.
  * Note that this is the same case as if checkpoint uploads would be broken and training proceeded past 250 steps.
* On requeue, job started training from "scratch", despite num_iterations proceeding from 6500 through ~7780. 
  * train.log also showed checkpoint saving for iteration 1250 (i.e. 6500+1250 = 7750) even though local saved checkpoints (in /mnt/scratch) were for step 7750.
  
This implies some mangled state between requeue / restore / checkpoint iteration count in logs vs what was saved. 

What _should have happened_ is that on requeue, given no checkpoints saved since starting from --restore-file, we should have restarted from --restore-file.

PR #143, opened 2022-06-07
Make Checkpoint Loading Strict

ArmenAg:
**Patch Description**
Currently if --restore-file is set, but there's failure in loading the file, we'll (somewhat) silently continue training with a randomly initialized model. This PR makes the failure explicit.

Specifically addresses #142

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -413,9 +413,7 @@ def load_checkpoint_to_cpu(path, arg_overrides=None, load_on_all_ranks=False) ->
             state = torch_load_cpu(local_path)
     except Exception as error:
         logger.error(
-            f"Got Exception While Trying To Load {path} with Paths to Load {paths_to_load}"
-        )
-        logger.error(
+            f"Got Exception While Trying To Load {path} with Paths to Load {paths_to_load}."
             "If you are not meaning to --restore-file, remove the command explictly."
         )
         raise error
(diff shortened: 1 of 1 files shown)


metaseq/checkpoint_utils.py
```diff
@@ -411,13 +411,15 @@ def load_checkpoint_to_cpu(path, arg_overrides=None, load_on_all_ranks=False) ->
...
-        print(
-            "got exception while trying to load",
-            path,
-            "with paths to load",
-            paths_to_load,
+    except Exception as error:
+        logger.error(
+            f"Got Exception While Trying To Load {path} with Paths to Load {paths_to_load}"
         )
+        logger.error(
```
suchenzang, line 418: Nit: single logger.error call with a newline?

metaseq/checkpoint_utils.py
```diff
@@ -411,13 +411,15 @@ def load_checkpoint_to_cpu(path, arg_overrides=None, load_on_all_ranks=False) ->
...
+            "If you are not meaning to --restore-file, remove the command explictly."
+        )
```
stephenroller, line 420:
```suggestion
        logger.error(
            f"Got Exception While Trying To Load {path} with Paths to Load {paths_to_load}."
            "If you are not meaning to --restore-file, remove the command explictly."
        )
```

stephenroller (approved):

PR #145, opened 2022-06-13
fix off by one in API

lilisierrayu:
**Patch Description**
The prompt_len in current code relies on the batch_generator implementation and can easily introduce error. Remove that dependence in this PR. 

**Testing steps**

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -627,12 +627,10 @@ class GeneratorInterface:
                     prompt_len = lengths[i]
                     if echo:
                         # don't cut off prompt
-                        tokens = tokens[: prompt_len + max_tokens[i] - 1]
-                        scores = scores[: prompt_len + max_tokens[i] - 1]
+                        tokens = tokens[: prompt_len + max_tokens[i]]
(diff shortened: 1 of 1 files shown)


metaseq/hub_utils.py
```diff
@@ -592,7 +592,9 @@ def generate(
 
             logger.info(f"Preparing generator with settings {self.cfg.generation}")
             generator = self.task.build_generator(
-                self.models, self.cfg.generation, extra_gen_cls_kwargs={"stop": stop}
+                self.models,
+                self.cfg.generation,
+                extra_gen_cls_kwargs={"stop": stop, "need_logprobs": True},
```
stephenroller, line 597: need log probs results in a lot more memory being used. we shouldn't use this unless we absolutely have to

lilisierrayu: That's reasonable. How about do: 
```
if logprobs > 0:
    generator.need_logprobs = True
```

lilisierrayu: Change applied.

stephenroller: Perfect

stephenroller (approved):
Approving conditional on change discussed

PR #148, opened 2022-06-13
Fix progress bar exception when both tensorboard and wandb are turned on

todpole3:
Change the progress bar creation logic so that only one of `WandBProgressBarWrapper` and `TensorboardProgressBarWrapper` is called. Calling these two wrappers sequentially will result in an exception.


stephenroller (approved):
thx!

PR #158, opened 2022-06-17
load last good checkpoint when corrupted checkpoint is passed

KUNAL1612:
**Patch Description**
- If checkpoint is corrupt (missing shards), load last good checkpoint in checkpoint directory
- If user passes custom checkpoint through restore_file, make sure checkpoint is not corrupt, else load last good checkpoint

**Testing steps**

- Manually removed shards from auto-generated last checkpoints 
- Passed non existent / corrupted checkpoints, model automatically loaded the last good checkpoint


metaseq/checkpoint_utils.py
```diff
@@ -179,14 +179,41 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
+def verify_shards(cfg, dir=None, checkpoint_name=None):
+    # verifies that all the shards of the checkpoint are present
+    checkpoint_name = checkpoint_name.replace(".pt", "")
+    num_gpus = int(cfg.save_dir[-1:])
```
qingliu1111, line 185: nit: Checkpoint naming format can be changing. So maybe it's better to read ngpu directly from config (cfg.distributed_world_size).

KUNAL1612: Hey Qing, the checkpoint config file doesn't seem to have a distributed_file_size key
image

KUNAL1612:
GPU tests do run on my local system but seem to be failing on circleci

image

KUNAL1612:
Had to make some edits to the way corrupted checkpoints passed by the user are handled when no last good checkpoint. Since trainer.load_checkpoint requires a string to be passed for the checkpoint_path_to_load parameter, it can't be None.

metaseq/checkpoint_utils.py
```diff
@@ -179,14 +179,41 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
+def verify_shards(cfg, dir=None, checkpoint_name=None):
+    # verifies that all the shards of the checkpoint are present
+    checkpoint_name = checkpoint_name.replace(".pt", "")
+    num_gpus = DistributedTrainingConfig.distributed_world_size
```
punitkoura, line 185: You should get the distributed_world_size from cfg instead... Is it not in cfg?

metaseq/checkpoint_utils.py
```diff
@@ -209,6 +236,31 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
             cfg.save_dir, "checkpoint_last{}.pt".format(suffix)
         )
         first_launch = not PathManager.exists(checkpoint_path_to_load)
+        if not first_launch and not verify_shards(
+            cfg, dir=cfg.save_dir, checkpoint_name=cfg.restore_file
+        ):
+            # checkpoint_last is corrupted
+            best_checkpoint = get_last_good_checkpoint(cfg)
```
punitkoura, line 243: Let's call get_last_good_checkpoint once instead of multiple times here

punitkoura (approved):
LGTM overall, left a few minor comments.

metaseq/checkpoint_utils.py
KUNAL1612: unfortunately, it is not in cfg
image

PR #163, opened 2022-06-20
Change constants to use environment variables when possible

xhluca:
This change allows the user to directly set different opt constants through environment variables rather than editing the constants.py file, which makes it easier when dynamically editing the files is not convenient (e.g. in docker).

### DIFF
--- a/metaseq/service/constants.py
+++ b/metaseq/service/constants.py
@@ -5,13 +5,13 @@
 
 import os
 
-MAX_SEQ_LEN = 2048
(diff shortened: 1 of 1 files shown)


stephenroller:
We likely will need to move to a more config based system, especially as I try to get the API to support serving multiple models.

xhluca:
Right, a `yml` file or something like that would make more sense here. The `os.getenv` I added serve as a stopgap as it only requires minimal changes to the `constants.py` file, as refactoring the code to use config files might require more effort.

PR #168, opened 2022-06-21
[api] Support temp=0. Return errors as json.

stephenroller:
**Patch Description**
- Makes the API return error messages as JSON objects consistent with the OpenAI api
- More robust to hangs derived from a bad user supplied arg
- Adds support for setting `temperature=0` to get greedy.

**Testing steps**
Deployed and using it with various test cases.

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -554,7 +554,6 @@ class GeneratorInterface:
         if temperature > 0:
             self.cfg.generation.temperature = temperature
         elif temperature == 0:
-            # temperature == 0
             self.cfg.generation.sampling = False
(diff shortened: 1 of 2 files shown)


metaseq/hub_utils.py
```diff
@@ -553,8 +553,12 @@ def generate(
         self.cfg.generation.beam = best_of
         if temperature > 0:
             self.cfg.generation.temperature = temperature
-        else:
+        elif temperature == 0:
+            # temperature == 0
```
suchenzang, line 557: Nit: remove?

suchenzang (approved):

PR #170, opened 2022-06-22
changes for interactive to work with non-fsdp

ngoyal2707:
changes for making interactive work without FSDP


klshuster (approved):

lilisierrayu:
this seems very helpful. Does this load with consolidated.pt too?

stephenroller (approved):
I feel like we keep writing many redundant variants of this

PR #181, opened 2022-06-24
Update convert_to_singleton.py

ArthurZucker:
Modified the consolidate part of the code to support the conversion of OPT66B.
@patrickvonplaten and I spent a bit of time debugging the script to get the restored.pt file for opt-66b. 
Thanks a lot @patrickvonplaten for your help. Here is what worked for us. 

Previous code would just overload the RAM (669 GB available were not enough) 
**Patch Description**
Changed `metaseq/scripts/convert_to_s`

**Testing steps**
Converted OPT-66B on 8 A100

Will try to test it one more time just to be sure


stephenroller (approved):
Cool, you all mind confirming the test before I merge?

PR #187, opened 2022-06-27
[api] Remove beam search

stephenroller:
**Patch Description**
After much consternation and fighting, I've greatly simplified our implementation of generation by eliminating beam search.

**Testing steps**
So much testing. But next PR will be unit tests.

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -620,7 +620,6 @@ class GeneratorInterface:
                 all_distributions = None
 
             # actually turn everything into strings
-            results = []
             for i in range(all_tokens.size(0)):
                 beams = []
                 for j in range(best_of):


metaseq/hub_utils.py
```diff
@@ -609,25 +609,25 @@ def generate(
...
+                for j in range(best_of):
                     # first beam is always the highest scoring
-                    tokens = beam["tokens"].tolist()  # implicit move to cpu
-                    scores = beam["positional_scores"].tolist()
-                    if logprobs > 0:
-                        distributions = beam["distributions"].cpu()
-                    else:
-                        distributions = None
-
+                    tokens = all_tokens[i, j].tolist()
```
klshuster, line 628: so `all_tokens` is `[bsz, n_best, n_tok]`?

klshuster (approved):
thanks for this. looks like a breaking change probably for some scripts (e.g. removing options) but should be straightforward

suchenzang (approved):
Oh good riddance - huge fan of simplification through deletion :P 

And glad lint is raising the same eyebrows here - just fix that nit?

PR #194, opened 2022-06-30
[cleanup] Delete SequenceScorer

stephenroller:
**Patch Description**
- SequenceScorer can be easily replicated via the API/hub_utils
- As a consequence delete it.
- As @tbmihailov has implemented the API-based evals, we don't need this code anymore.

**Testing Steps**
CI. Only code was deleted.


tbmihailov:
Stephen, let me check if this is not used by our other experiments! We could move this to the eval code if it is used but lets not break it just yet!

tbmihailov:
> Stephen, let me check if this is not used by our other experiments! We could move this to the eval code if it is used but lets not break it just yet!

We are still using the SequenceScorer for scoring in the multi-choice setting (where prompts tokens are scored). What is the new way that this works in the API that it is not needed anymore? We could probably switch to the new code-path?

stephenroller:
You give echo=True to the API

suchenzang:
@tbmihailov have you had a chance to check if switching to the API with echo=True works for your evals?

tbmihailov:
> @tbmihailov have you had a chance to check if switching to the API with echo=True works for your evals?

It would not work right now because it supposes that we have an API running. Which is very tedious to do for each checkpoint. For example for the XLMG experiments (that we still need to finish), we are evaluating the model every X steps. Having the ability to run locally is essential! Lets rename this to SequenceScorerLegacy or something if the presence is disturbing?

suchenzang:
> Lets rename this to SequenceScorerLegacy or something if the presence is disturbing? 
Ah no need - will just wait for whenever your evals are done!

PR #44, opened 2022-05-05
[Bug Fixed] Add quote for `train_cmd` in `metaseq/launcher/slurm.py`

wkcn:
Hi there, thanks for your great work!
I found that there is no quote for `train_cmd` in `metaseq/launcher/slurm.py`.
It leads to an issue that `dry-run` prints a command without quote, namely `--adam-betas (0.9, 0.95)`, which could not run directly. The correct command should be `--adam-betas '(0.9, 0.95)'`.

**Patch Description**
Add quote for `train_cmd` in the function `gen_train_command` of `metaseq/launcher/slurm.py`.
The command `--adam-betas (0.9, 0.95)` will be fixed to `--adam-betas '(0.9, 0.95)'` in this PR : )

**Testing steps**
Run the following command to print the `dry-run` message.
```bash
opt-baselines \
  -n 1 -g 2 \
  -p test_v0 \
  --model-size 125m \
  --azure \
  --data ./dataset \
  --checkpoints-dir "./outputs" \
  --local \
  --dry-run
```

PR #195, opened 2022-06-30
add integration test to make sure training is successful

KUNAL1612:
**Patch Description**
Added a small gpu test to ensure training works as expected. A small dataset (~18MB) is downloaded from a public S3 bucket and it is trained for 5 epochs. The files are then cleaned up.

**Testing steps**
Launched a slurm job with 8 GPU's to test this. Compared returned loss value with expected loss value after 5 updates/5 epochs. Implemented PR #44 and some other changes to make sure that training broke and the assertions in the test failed. Also tried running in non-GPU environment to make sure test failed there too.


metaseq/checkpoint_utils.py
```diff
@@ -179,14 +179,41 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
+def verify_shards(cfg, dir=None, checkpoint_name=None):
```
punitkoura, line 182: @KUNAL1612 This checkpoint verification code should already be checked in by now right? Why is this a part of this PR?

stephenroller:
Super cool. Let's talk about this internally.

metaseq/checkpoint_utils.py
KUNAL1612: I think because no one approved the previous PR so I couldn't merge it. I re-sent requests but am still waiting for approval

suchenzang: Which previous PR is needed?

suchenzang:
There's internal-specific logic getting checked into here, along with duplicate copies of files. Will need both to get cleaned up (unless duplication is unavoidable for some reason).

metaseq/checkpoint_utils.py
KUNAL1612: #158 but I just merged it

stephenroller:
Yeah I wonder if it wouldn't be better to tarball up the data, put it on s3, and download it at the start of jobs. I don't love checking data into repos.

suchenzang:
The vocab related files were already checked in for the OPT release: https://github.com/facebookresearch/metaseq/tree/main/projects/OPT/assets

KUNAL1612:
> Yeah I wonder if it wouldn't be better to tarball up the data, put it on s3, and download it at the start of jobs. I don't love checking data into repos.

@stephenroller How would you suggest I go about uploading and then using tools to download it? More specifically, where do I upload the data and what are the tools I need to download them during the test

stephenroller:
@punitkoura should be able to help upload it to FAIR's s3 bucket

KUNAL1612:
Removed the files from the PR and now loading them from the FAIR S3 public bucket

PR #205, opened 2022-07-05
[api] API keys now required

stephenroller:
**Patch Description**
Previously, all API keys (including empty) would be accepted. Now we throw an authorization error if the user fails to provide one.

Also denylists a few placeholders that are in internal/external docs.

**Testing steps**
Manual testing

### DIFF
--- a/metaseq_cli/interactive_hosted.py
+++ b/metaseq_cli/interactive_hosted.py
@@ -187,20 +187,33 @@ def handle_exception(e):
     # pass through HTTP errors
     if isinstance(e, HTTPException):
         return e
-    # now you're handling non-HTTP exceptions only
-    response = jsonify(
-        {
-            "error": {
-                "message": str(e),
-                "type": "oops",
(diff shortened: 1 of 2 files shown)


metaseq_cli/interactive_hosted.py
```diff
@@ -187,20 +187,36 @@ def handle_exception(e):
...
+    # denylist a few placeholders various people have used
+    if key == "":
+        return False
+    if "YOUR_NAME_HERE" in key:
+        return False
+    if "$USER" in key:
+        return False
+    if "your-key-here" in key:
+        return False
+    return True
```
suchenzang, line 207: Wait this is all we do for validation? lol

suchenzang (approved):
Just a q on validation logic. Will leave up to you on how auth will work going forward.

metaseq_cli/interactive_hosted.py
stephenroller: before we did nothing :P

PR #212, opened 2022-07-08
Better validation in JSONL datasets

KUNAL1612:
**Patch Description**
Added code to check for validity of json in jsonl files in the index building function. Also added two new tests to make sure that empty jsonl files as well as corrupt jsonl files throw up the right errors.

**Testing steps**
ran the new tests to ensure that it works as expected

### DIFF
--- a/metaseq/data/jsonl_dataset.py
+++ b/metaseq/data/jsonl_dataset.py
@@ -15,7 +15,7 @@ from typing import Callable, Optional
 
 import numpy as np
 import torch
-from metaseq import pdb
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


metaseq/data/jsonl_dataset.py
```diff
@@ -88,6 +88,13 @@ def _build_index(self, path: str):
         cur = 0
         while True:
             line = f.readline()
+            if line != b"":
```
stephenroller, line 91: why is this bytes?

metaseq/data/jsonl_dataset.py
```diff
@@ -88,6 +88,13 @@ def _build_index(self, path: str):
...
+                try:
+                    json.loads(line)
+                except json.decoder.JSONDecodeError:
+                    raise AssertionError(
+                        "Unable to load truncated or corrupted JSON file"
```
stephenroller, line 96: Instead of an assertion error, let's keep with JSONDecodeError but provide a better error message. Include the name of the file and the line number in the error message, please

stephenroller:
Great tests!

metaseq/data/jsonl_dataset.py
KUNAL1612: I noticed that that's just how line was formatted. And if line was b"" then it would always fail json loading anyway

metaseq/data/jsonl_dataset.py
KUNAL1612: Noted. Updated the file to do this.

punitkoura (approved):

PR #214, opened 2022-07-11
Convert to singleton bug fix + Test for singleton model compatibility between metaseq and HF

punitkoura:
**Patch Description**
Fixing a bug in the convert to singleton script, and adding a test to verify that the conversion script works as intended, and the model outputs from a 125m singleton checkpoint are the same on metaseq and HF

**Testing steps**
Added integration tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,9 @@ install_dep_common: &install_dep_common
         pip install -i https://test.pypi.org/simple/ bitsandbytes-cuda111 -U
         python -c 'import torch; print("Torch version:", torch.__version__)'
         python -m torch.utils.collect_env
+        # Need to install ninja build system
+        sudo apt-get update
+        sudo apt-get install ninja-build
 
 install_dep_fused_ops: &install_dep_fused_ops
   - run:
@@ -46,7 +49,7 @@ install_dep_fused_ops: &install_dep_fused_ops
           git clone --depth=1 --branch fairseq_v2 https://github.com/ngoyal2707/Megatron-LM.git
           cd Megatron-LM
           pip install -r requirements.txt
-          pip install .
+          pip install -e .
           cd ~/
         fi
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -132,6 +135,19 @@ create_conda_env: &create_conda_env
         python --version
         pip install --upgrade pip
 
+download_and_configure_125m_with_hf_dependencies: &download_and_configure_125m_with_hf_dependencies
+  - run:
+      name: Download and configure a 125m checkpoint with HF dependencies
+      working_directory: ~/metaseq/gpu_tests
+      command: |
+        source activate metaseq
+        if [[ $(python -c "import torch; print(torch.__version__)") = "1.9.1+cu111" ]]; then
```
punitkoura, line 144: The metaseq model loading doesn't seem to work well with pytorch 1.8, so only running the test for pt19

punitkoura:
@patrickvonplaten Could you also have a look at this test? I have adapted the script you had created to test compatibility between singleton megatron checkpoint and HuggingFace. Thank you!

.circleci/config.yml
stephenroller: We can kill pytorch 1.8. It's behind and internally we're using 1.10

stephenroller: Let's create an issue for that for the future.

gpu_tests/test_hf_compatibility.py
```diff
@@ -0,0 +1,194 @@
...
+
+def setup_vocab_and_merges(model_path):
+    vocab_file = os.path.join(model_path, "gpt2-vocab.json")
+    merges_file = os.path.join(model_path, "gpt2-merges.txt")
+    tokenizer = GPT2Tokenizer(vocab_file, merges_file)
+    tokenizer.save_pretrained(model_path)
+    return vocab_file, merges_file, tokenizer
+
+
+def load_mp_model_and_run_eval(cfg: MetaseqConfig, **kwargs):
```
stephenroller, line 47: Should these things be moved to hubutils or similar?

gpu_tests/test_hf_compatibility.py
```diff
@@ -0,0 +1,194 @@
...
+        logits[:orig_dim].unsqueeze(0)
+        for logits, orig_dim in zip(gathered_logits, orig_dims)
+    ]
+
+    for index, logits in enumerate(trimmed_logits):
+        torch.save(logits, f"/tmp/test_hf_compatibility_{index}.pt")
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "test requires a GPU")
+@unittest.skipIf(
```
stephenroller, line 113: given that you have the skip here, let's maybe not carve out the special case in the bash setup?

gpu_tests/test_hf_compatibility.py
```diff
@@ -0,0 +1,194 @@
...
+    version.parse(torch.__version__) < version.parse("1.9.1"),
+    "test requires a pytorch version of at least 1.9.1",
+)
+class TestTraining(unittest.TestCase):
```
stephenroller, line 117: why is this called TestTraining?

stephenroller (approved):
great tests

gpu_tests/test_hf_compatibility.py
punitkoura: Moved the common functions to hub_utils.py

I've kept load_mp_model_and_run_eval in this file still, since the logic there is very specific to this test... And we are using common functions in the function body (like fsdp_enable_wrap, or checkpoint_utils.load_model_ensemble_and_task).

gpu_tests/test_hf_compatibility.py
punitkoura: Good point... The problem is that we need pytorch 1.9 to run the convert to singleton script as well, which runs in bash, not in the test... Hence I had to add a separate clause in bash.

PR #224, opened 2022-07-15
[api] Fix logit outputs

stephenroller:
**Patch Description**
@tbmihailov pointed out that our logits differ from the openai's, in that we (previously) started our logprobs with the first generated token, rather than including a sentinel `null` logprob and including the first token.

This PR fixes that so we are more consistent with OpenAI

**Testing steps**
So much manual testing.

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -647,22 +647,30 @@ class GeneratorInterface:
                         pass
                     else:
                         # cut off prompt
-                        tokens = tokens[prompt_len:][: max_tokens[i]]
-                        scores = scores[prompt_len:][: max_tokens[i]]
+                        print_r0("tokens after prompt_len:", tokens[prompt_len + 1 :])
+                        print_r0("scores after prompt_len:", scores[prompt_len + 1 :])
+                        tokens = tokens[prompt_len + 1 :][: max_tokens[i]]
+                        scores = scores[prompt_len + 1 :][: max_tokens[i]]
                         if logprobs > 0:
-                            distributions = distributions[prompt_len:][: max_tokens[i]]
+                            distributions = distributions[prompt_len + 1 :][
+                                : max_tokens[i]
+                            ]
+
+                    # cut off the starting token
+                    tokens_no_eos = tokens[1:] if echo else tokens
+                    scores_with_eos = [None] + scores[1:] if echo else scores
                     # turn it into a string
-                    text = self.bpe.bpe.decode(tokens)
+                    text = self.bpe.bpe.decode(tokens_no_eos)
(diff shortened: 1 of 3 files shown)


metaseq/sequence_generator.py
```diff
@@ -141,27 +141,23 @@ def _generate(
 
         # initialize buffers
         scores = (
-            torch.zeros(bsz * beam_size, max_len + 1).to(src_tokens).float()
+            torch.zeros(bsz * beam_size, max_len).to(src_tokens).float()
         )  # +1 for eos; pad is never chosen for scoring
```
suchenzang, line 145: Does that +1 comment for eos still apply?

metaseq/sequence_generator.py
```diff
@@ -254,8 +254,6 @@ def _generate(
                 model_out, log_probs=True, sample=None
             )
             lprobs = lprobs[:, -1, :]
-            if self.need_logprobs:
```
suchenzang, line 257: Conditionally seem to be missing the branch of `step >= max_len and self.need_logprobs == True` now - is that intended?

metaseq/hub_utils.py
```diff
@@ -700,23 +711,22 @@ def _filter_special(
...
         output = []
         mask = []
-        for t, s in zip(tokens, scores):
+        for i, (t, s) in enumerate(zip(tokens, scores)):
             if t == pad_token:
                 # simply skip pads
                 mask.append(False)
                 continue
-            if t <= 3:
+            if t <= 3 and i > 0:
```
suchenzang, line 721: This seems brittle, but probably making it less brittle requires too much glue.  We have asserts to make sure we're using the matching dictionary in streaming lm: https://github.com/facebookresearch/metaseq/blob/d2c75c2ca38c5a68e18c6c4aee2e06e78e2d7ab5/metaseq/tasks/streaming_language_modeling.py#L158-L166 - can we do something similar here?

suchenzang (approved):

metaseq/hub_utils.py
stephenroller: nvm i just did the right thing

metaseq/sequence_generator.py
stephenroller: yes. the step == max_len step is forced to be an EOS token regardless. So we don't need to record its score into the logprobs (that's also why scores is protected from writing)

dongs0104:
hello @stephenroller i have question about this,
 when i using huggingface model, is it require to fix or not?
Thanks

stephenroller:
If you're using HF Transformers, then it's their own generation code.

PR #225, opened 2022-07-15
[api] Add a clutch sync

stephenroller:
**Patch Description**
Add a very important sync so that we don't worry about queuing future generations onto the GPU before old generations are done.

**Testing steps**
Looooots of debugging with @klshuster and @ngoyal2707.

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -678,6 +678,7 @@ class GeneratorInterface:
                     beams.append(result)
                 retval.append(beams)
 
+        torch.cuda.synchronize()
         logger.info(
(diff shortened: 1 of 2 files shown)


metaseq/hub_utils.py
```diff
@@ -678,6 +678,7 @@ def generate(
                     beams.append(result)
                 retval.append(beams)
 
+        torch.cuda.synchronize()
```
suchenzang, line 681: Nit: add a comment for why this sync was added / what happens to generated output if it gets removed?

suchenzang (approved):

PR #226, opened 2022-07-15
[api] Fix bad batching causing wrong number of tokens returned or cutoff.

stephenroller:
**Patch Description**
There are some adversarial setups where a user could produce a long prompt (2040 prompt, 8 gen) and another user could produce a very short prompt with long generation (8 prompt, 2040 gen). These two might get batched together because individually they're below 2048. However, once padded, it becomes an input of 4080. This can cause one of the two outputs to be returned incorrect.y

**Testing steps**
Manual testing


metaseq_cli/interactive_hosted.py
```diff
@@ -89,7 +89,12 @@ def batching_loop(timeout=100, max_tokens=MAX_BATCH_TOKENS):
             # accumulate the batch until it gets too big
             longest = max([item] + batch).cost
             batch_cost = longest * (len(batch) + 1)
-            if batch and batch_cost > max_tokens:
+            # overflow corresponds to whether max(prompt_len) + gen_len will
+            # fit the max sequence length
+            max_prompt_len = max(x.prompt_len for x in [item] + batch)
+            max_gen_len = max(x.gen_len for x in [item] + batch)
+            overflow = max_prompt_len + max_gen_len < MAX_SEQ_LEN
+            if batch and batch_cost > max_tokens and overflow:
```
suchenzang, line 97: Should this be `batch and batch_cost > max_tokens or overflow` instead of `and overflow`?

stephenroller: Yup. Actually caught this last night and patched it, but didn't push the updated diff.

suchenzang (approved):

PR #227, opened 2022-07-15
[api] Prevent unlike generations from being grouped.

stephenroller:
**Patch Description**
Bug in the queuing logic means that we actually are grouping generations that we shouldn't.

**Testing steps**
Manual testing.


suchenzang:
Hmm I don't see how this addresses the grouping issue, but I'm guessing you've tested it in conjunction with the overflow PR? (don't see the diff from https://github.com/facebookresearch/metaseq/pull/226 in here...)

stephenroller:
This PR is about making sure we don't mix temperatures or nucleuses in the same batch. The logic for that already existed (these target queues are already stratified by those parameters). This bug was just that we were resetting which queue we pulled from every time.

PR #197, opened 2022-07-03
[Cleanup]: remove upgrade_state_dict logic, normformer, unused alignment/encoder args

suchenzang:
Tackling https://github.com/facebookresearch/metaseq/issues/165

Only checkpoints to work with here are with OPT models so far.

More cleanup around unused args / logic in preparation for reducing complexity around configuration / checkpointing logic.

* Removed all normformer args & related classes (sync ln)
* Removed stable embeddings 
* Removed FeedForwardNetwork class

PR #229, opened 2022-07-16
Remove upgrade_state_dict logic

suchenzang:
No longer needed for backwards compatibility with pre-OPT days.

See https://github.com/facebookresearch/metaseq/pull/197#issuecomment-1173104694


suchenzang:
Good work on https://github.com/facebookresearch/metaseq/pull/214 @punitkoura ! This new test was able to catch a regression here :)

stephenroller (approved):

PR #231, opened 2022-07-17
[Cleanup] Delete unused methods / imports / args / classes: 2/N

suchenzang:
Following https://github.com/facebookresearch/metaseq/pull/230 and continuing to break down https://github.com/facebookresearch/metaseq/pull/197.

Removed:
* unused `moe_disable_padding` arg
* unused `from_pretrained` methods, since we currently depend on `load_model_ensemble_and_task` from `checkpoint_utils` (not great but saving that for another PR)
* unused `hub_models` method
* `EncoderDecoderModel` class


metaseq/models/base_model.py
```diff
@@ -168,131 +168,6 @@ def apply_prepare_for_onnx_export_(module):
...
-            moe_disable_padding=moe_disable_padding,
-            skip_prepare_for_inference=skip_prepare_for_inference,
-        )
-
-    @classmethod
-    def hub_models(cls):
-        return {}
-
-
-class EncoderDecoderModel(BaseModel):
```
stephenroller, line 220: please break this up further, separating the changes to hub utils from the removal of enc dec. I put it at 50% that we're going to want enc dec back in the future and i'd like to make that easier to undo.

stephenroller: also, can this be moved to be on top of main instead of a stacked PR?

suchenzang: Yes, already done. Part 1: https://github.com/facebookresearch/metaseq/pull/244

Issue #220, opened 2022-07-13
Create generation benchmarks on different model sizes


punitkoura:
## 🚀 Feature Request
The goal of this task is to track the generation speed (in WPS) of the various OPT model sizes. We can also include profiling to identify potential bottlenecks in the generation process. This investigation would help us speed up generation.

### Motivation
To improve the speed of text generation from OPT models.

### Pitch
Create a generation benchmark that takes in any of our models, and runs a fixed generation, and reports timing per token. Model name should be configurable.

### Alternatives
N/A

### Additional context
N/A

KUNAL1612:
Addressed in #240

PR #240, opened 2022-07-21
Add script to measure generation benchmarks for different models sizes

KUNAL1612:
**Patch Description**
Added a script that takes in one/many model sizes as input and outputs generation statistics. Addresses #220. Also allows for custom data to be passed as input. 

**Testing steps**
Run this with multiple model sizes to observe WPS counts. Can take multiple models as input and report output for each.

Usage : 

` python generation_benchmarks.py --model_name 2.7b 13b --padding_size 32`

NB: model checkpoint sizes for 66b are huge and the script will require sufficient storage and time tor run.

### DIFF
--- a/metaseq/scripts/generation_benchmarks.py
+++ b/metaseq/scripts/generation_benchmarks.py
@@ -233,10 +233,13 @@ def download_prompts():
     pass
 
 
-def cleanup(prompts):
+def cleanup_data():
     shutil.rmtree("./dependencies")
     os.remove("./dependencies.tar.gz")
-    os.remove(prompts)
+
+def cleanup_prompts(prompt_path):
+    os.remove(prompt_path)
+
 
 
 if __name__ == "__main__":
@@ -245,6 +248,9 @@ if __name__ == "__main__":
     parser.add_argument("--prompt_path", nargs="?")
     parser.add_argument("--padding_size", nargs="?", default="512")
     args = parser.parse_args()
+
+    if args.model_name is None:
+        raise AssertionError("Please pass at least one model name as argument")
 
     if args.prompt_path is None:
         # using default prompts
@@ -268,4 +274,6 @@ if __name__ == "__main__":
                 model_path="./dependencies",
                 padding_size=int(args.padding_size),
             )
-        cleanup(args.prompt_path)
+        cleanup_data()
+
+    cleanup_prompts(prompt_path = args.prompt_path)
--- a/metaseq/scripts/generation_benchmarks.py
+++ b/metaseq/scripts/generation_benchmarks.py
@@ -1,0 +1,291 @@
+import os
+import json
+import time
+from transformers import GPT2Tokenizer
+from metaseq import checkpoint_utils, tasks, utils
+from transformers import OPTForCausalLM
+from packaging import version
+import torch
+import unittest
+import torch.nn.functional as F
+from metaseq.scripts.convert_to_singleton import create_generation_config_with_defaults
+from metaseq.distributed import utils as dist_utils
+from metaseq.distributed import fsdp_enable_wrap, fsdp_wrap
+from metaseq.dataclass.configs import MetaseqConfig
+from metaseq import pdb
+import argparse
+import urllib.request
+import tarfile
+import shutil
+
+prompts = []
+
+
+def load_prompts(path, max_len=10, max_sentences=20):
+    all_data = list(open(path, "r"))
+    for data in all_data:
+        json_data = json.loads(data)
+        s = " ".join(json_data["text"].split()[:max_len])
+        prompts.append(s)
+        if len(prompts) == max_sentences:
+            break
+
+
+def setup_vocab_and_merges(model_path):
+    vocab_file = os.path.join(model_path, "gpt2-vocab.json")
+    merges_file = os.path.join(model_path, "gpt2-merges.txt")
+    tokenizer = GPT2Tokenizer(vocab_file, merges_file)
+    tokenizer.save_pretrained(model_path)
+    return vocab_file, merges_file, tokenizer
+
+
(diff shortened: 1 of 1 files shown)


metaseq/scripts/generation_benchmarks.py
```diff
@@ -0,0 +1,279 @@
...
+
+def download_data(key):
+    links_to_data = {}
+    links_to_data["125m"] = [
+        "https://dl.fbaipublicfiles.com/opt/v1_20220502/125m/reshard-model_part-0.pt",
+        "https://dl.fbaipublicfiles.com/opt/v1_20220502/125m/reshard-model_part-1.pt",
+    ]
+    links_to_data["350m"] = [
+        "https://dl.fbaipublicfiles.com/opt/v1_20220502/350m/reshard.pt"
+    ]
```
stephenroller, line 172: Let's drop this one because it's a special snowflake

metaseq/scripts/generation_benchmarks.py
```diff
@@ -0,0 +1,279 @@
...
+    with torch.no_grad():
+        model(prompt_ids)[0]
+
+    end = time.time()
+    times_list = torch.tensor(end - begin + prompt_loading).cuda()
+
+    torch.distributed.all_gather(
+        thread_times, times_list, group=dist_utils.get_global_group()
+    )
+    torch.save(thread_times, "./dependencies/times.pt")
```
stephenroller, line 96: you don't need to save an external file. you can pass return values up from the distributed.call_main function.

metaseq/scripts/generation_benchmarks.py
```diff
@@ -0,0 +1,279 @@
...
+        file = tarfile.open("./dependencies.tar.gz")
+        file.extractall()
+
+    # download model checkpoint
+    for shard in links_to_data[key]:
+        file_name = shard.split("/")[-1]
+        urllib.request.urlretrieve(shard, "./dependencies/" + file_name)
+
+
+def download_prompts():
```
stephenroller, line 230: instead of downloading prompts (and cleaning them up), I suggest we just write 10 (short) prompts and keep them as constants at the top of the file.

metaseq/scripts/generation_benchmarks.py
```diff
@@ -0,0 +1,279 @@
...
+
+def cleanup_prompts(prompt_path):
+    os.remove(prompt_path)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--model_name", nargs="+")
+    parser.add_argument("--prompt_path", nargs="?")
+    parser.add_argument("--padding_size", nargs="?", default="512")
```
stephenroller, line 249: can padding be automatically computed from the inputs?

stephenroller (commented):
Wow awesome to see this right before my weekend!  Let's try to simplify it and remove the dependencies. @ngoyal2707 may also have suggestions on how we actually implement timing.

We may also wish to switch to using sequence generator so that we can generate N tokens instead of just scoring.

metaseq/scripts/generation_benchmarks.py
KUNAL1612: Fair enough for 10 prompts but I was wondering if someone wanted more (the file has up to 1000) then it may just be too much clutter.  Also isn't it generally better to average out numbers from more prompts than from just 10?

metaseq/scripts/generation_benchmarks.py
KUNAL1612: Yes as in padding = length of longest sentence. But in the same spirit as having more sentences, having a higher padding size would mean more generated words and hence a more true average?

metaseq/scripts/generation_benchmarks.py
```diff
@@ -0,0 +1,279 @@
...
+        model = models[0]
+
+    model.summon_full_params()
+    model = model.eval()
+
+    begin = time.time()
+    with torch.no_grad():
+        model(prompt_ids)[0]
+
+    end = time.time()
```
ngoyal2707, line 90: https://discuss.pytorch.org/t/how-to-measure-time-in-pytorch/26964 is more correct way to measure pytorch / cuda code timing.

your current way doesn't ensure that the work submitted to cuda stream by `model(prompt_ids)[0]` is completed by the time you get to line 90

ngoyal2707 (changes requested):
putting a blocking comment for changes needed for synchronization needed for timing the code

metaseq/scripts/generation_benchmarks.py
KUNAL1612: I see. Thanks for the input!

metaseq/scripts/generation_benchmarks.py
stephenroller: If all the prompts are the same length, I don't expect any variance at all. Really I think it's okay if we just have 1 prompt, and averaged over many runs of the same prompt.

metaseq/scripts/generation_benchmarks.py
stephenroller: In practice, our batching tries to group similarly lengthed examples together to minimize padding. So some padding is realistic, but we actively try to minimize it.

KUNAL1612: hmm I see. So if I have one sentence (as per comment above) and have that run, say, 1000 times, will the padding being equal to the sentence length make sense? In which case padding will no longer be an argument

KUNAL1612:
> putting a blocking comment for changes needed for synchronization needed for timing the code

Addressed these issues

punitkoura (approved):
Approving since all change requests have been added.

ngoyal2707 (approved):
didnt review thoroughly but the timing part looks correct to me now

PR #242, opened 2022-07-22
[api] Cut off user's prompt

stephenroller:
**Patch Description**
This patch shortens a user's prompt to meet the number of generated tokens they request.

**Testing steps**
Deployed API


metaseq_cli/interactive_hosted.py
```diff
@@ -298,16 +298,20 @@ def completions(engine=None):
 
     ret_queue = queue.Queue()
     for i, prompt in enumerate(prompts):
+        gen_len = generation_args.get("max_tokens", 0)
+        if gen_len + len(prompt) + 1 > MAX_SEQ_LEN:
+            # cut off the prompt to always fit with number of generations we need
+            prompt = prompt[-(MAX_SEQ_LEN - gen_len - 1) :]
+        prompt = prompt[-(MAX_SEQ_LEN - gen_len) :]
```
suchenzang, line 305: Uh.... maybe my brain is dead but:

Say `MAX_SEQ_LEN - gen_len` == 5. For L304 to execute, `len(prompt) + 1 > MAX_SEQ_LEN - gen_len`, so `len(prompt) > 4`.  Then `prompt[-(MAX_SEQ_LEN - gen_len - 1) :]` == `prompt[-4:]` which takes the last four items of prompt.

In L305, you try taking the last 5 elements of `prompt`. So if `len(prompt) ==  6` or something, you're going to be out of bounds in L305 right?  Or is that never possible?

stephenroller: you're probably right

stephenroller: I think i shouldn't have both these lines... but which one...

stephenroller: got it

suchenzang (approved):

PR #244, opened 2022-07-22
[Cleanup] Delete unused methods / imports / args / classes: 2.1/N

suchenzang:
Splitting https://github.com/facebookresearch/metaseq/pull/231 up into 2 PRs.

Removed:
* unused `moe_disable_padding` arg
* unused `from_pretrained` methods, since we currently depend on `load_model_ensemble_and_task` from `checkpoint_utils` (not great but saving that for another PR)
* unused `hub_models` method


stephenroller:
I guess what i'm confused about is why this PR kills encoder_states instead of moving that to the enc-dec deletion PR. We're having some sort of miscommunication here.

suchenzang:
@stephenroller if that's the one that bugs you, I'll revert that one commit.  Reason this isn't tied to enc-dec deletion is because it originated from deleting an unused flag that gates the entire codepath related to encoder_states.

stephenroller (approved):

stephenroller:
Thanks for the clarification

stephenroller:
(The reason being that if we do bring back enc dec, the generator will need access to those states usually)

PR #245, opened 2022-07-22
[checkpoint] Fix race condition with blob.

stephenroller:
**Patch Description**
When downloading checkpoints from blob, we have a race condition where not all checkpoints may have finished downloading to local workers, thus causing an exception. This ensures everyone patiently waits on everyone else.

**Testing steps**
cm3 branch


suchenzang (approved):

metaseq/checkpoint_utils.py
```diff
@@ -381,6 +381,11 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
             checkpoint_path_to_load = save_dir_last
 
     logger.info(f"attempting to load checkpoint from: {checkpoint_path_to_load}")
+    import torch.distributed
+    import metaseq.distributed.utils as dist_utils
```
suchenzang, line 385: Wait why are you re-importing this?

PR #247, opened 2022-07-22
[api] Keep logits in fp32

stephenroller:
**Patch Description**
This patch keeps our logit scores in fp32 so they are more accurately reported to the user. Has the nice added benefit of fixing a known bug where using temperature 0.1 causes overflow issues.

**Testing steps**
@tbmihailov running evals on this

### DIFF
--- a/metaseq/sequence_generator.py
+++ b/metaseq/sequence_generator.py
@@ -179,7 +179,8 @@ class SequenceGenerator(nn.Module):
         )
         # normalize
         model_predictions = model_out[0].float()
-        model_predictions.div_(self.temperature)
+        if self.temperature > 0 and self.temperature != 1.0:
+            model_predictions.div_(self.temperature)
         # lprobs is the log probability of each possible token in every position
         # lprobs \in FloatTensor(bsz * beam_size, prompt_len, vocab_size)
         lprobs = self.model.get_normalized_probs(model_predictions, log_probs=True)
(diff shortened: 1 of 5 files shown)


metaseq/sequence_generator.py
```diff
@@ -249,8 +250,9 @@ def _generate(
                 tokens[:, : step + 1],
                 incremental_state=incremental_states,
             )
-            model_out[0].div_(self.temperature)
-            lprobs = self.model.get_normalized_probs(model_out, log_probs=True)
+            model_predictions = model_out[0].float()
```
suchenzang, line 253: Add a comment here for which bug this fixes? Don't want future eager optimizers to get excited about taking this out...

suchenzang (approved):

metaseq/sequence_scorer.py
```diff
@@ -72,7 +72,7 @@ def gather_target_probs(probs, target):
             for bd, tgt, is_single in batched:
                 sample["target"] = tgt
                 curr_prob = model.get_normalized_probs(
-                    bd, log_probs=len(models) == 1
+                    bd[0], log_probs=len(models) == 1
```
lilisierrayu, line 75: why not pass  bd[0].float() to be consistent with generator?

metaseq/sequence_generator.py
stephenroller: we're currently in a live debugging session lmao

PR #248, opened 2022-07-22
Fix tokenizer import

punitkoura:
**Patch Description**
Fix tokenizer import to be from tokenizers instead of transformers to be in line with https://github.com/facebookresearch/metaseq/blob/main/metaseq/tasks/streaming_language_modeling.py#L127

**Testing steps**
CI tests


stephenroller (approved):
Ty for the speedy fix!

PR #249, opened 2022-07-22
[Cleanup] Delete unused methods / imports / args / classes - EncoderDecoder: 2.2/N

suchenzang:
Second half of https://github.com/facebookresearch/metaseq/pull/231

* `EncoderDecoderModel` class


stephenroller (approved):
thx

PR #256, opened 2022-07-26
[Cleanup] Remove deprecated Trainer.get_meter: 8/N

suchenzang:
See title. Not called anywhere (all replaced by metrics.get_meter(s))


stephenroller (approved):
are we sure this isn't used?

PR #261, opened 2022-07-27
Fix Bug in Loading Checkpoint Post load_state_dict Refactor

ArmenAg:
**Patch Description**
Fix a small nit in the load_state_dict refactor code

**Testing steps**
I ran training, that loaded checkpoint with this small change turned on. Everything worked.


suchenzang (approved):
not a nit - this is an actual bug i put in 😅. thanks for catching and fixing!!

PR #263, opened 2022-07-27
import tasks, models and criterions from metaseq-internal if exists.

urielsinger:
Draft PR for structure approval (together with PR #333 in metaseq-internal).

### DIFF
--- a/metaseq/criterions/__init__.py
+++ b/metaseq/criterions/__init__.py
@@ -31,3 +31,8 @@ for file in os.listdir(os.path.dirname(__file__)):
     if file.endswith(".py") and not file.startswith("_"):
         file_name = file[: file.find(".py")]
         importlib.import_module("metaseq.criterions." + file_name)
+
+for file in os.listdir(os.path.dirname(__file__).replace('metaseq/metaseq', 'metaseq-internal/metaseq_internal')):
+    if file.endswith(".py") and not file.startswith("_"):
+        file_name = file[: file.find(".py")]
+        importlib.import_module("metaseq_internal.criterions." + file_name)
--- a/metaseq/criterions/__init__.py
+++ b/metaseq/criterions/__init__.py
@@ -32,7 +32,9 @@ for file in os.listdir(os.path.dirname(__file__)):
         file_name = file[: file.find(".py")]
         importlib.import_module("metaseq.criterions." + file_name)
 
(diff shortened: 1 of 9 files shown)


metaseq/trainer.py
```diff
@@ -1002,7 +1002,7 @@ def clip_grad_norm(
             clip_norm,
             clip_norm_type,
             aggregate_norm_fn=None,
-            skip_gradient_update_on_clip_norm=skip_gradient_update_on_clip_norm,
+            # skip_gradient_update_on_clip_norm=skip_gradient_update_on_clip_norm,
```
suchenzang, line 1005: Nit: probably don't want to hardcode this commented out flag.

metaseq/criterions/__init__.py
```diff
@@ -31,3 +31,10 @@ def build_criterion(cfg: DictConfig, task):
     if file.endswith(".py") and not file.startswith("_"):
         file_name = file[: file.find(".py")]
         importlib.import_module("metaseq.criterions." + file_name)
+
+criterions_dir = os.path.dirname(__file__).replace('metaseq/metaseq', 'metaseq-internal/metaseq_internal')
```
stephenroller, line 35: What if we just did:

```
try:
    import metaseq_internal.criterions
except ImportError:
    pass
```

And we move all this dynamic importing into `metaseq_internal.criterions.__init__.py`?

Ditto for other approaches.

setup.py
```diff
@@ -168,6 +168,12 @@ def do_setup(package_data):
             "torch",
             "tqdm",
             "typing_extensions",
+            "wandb",
+            "webdataset==0.1.103",
+            "glide_text2im @ git+https://github.com/openai/glide-text2im",
+            "clip @ git+https://github.com/openai/CLIP.git",
+            "matplotlib",
+            "opencv-python"
```
stephenroller, line 176: could we move these into the metaseq internal setup.py instead for now?

metaseq/trainer.py
urielsinger: i'll remove it entirely as this is an unknown argument

metaseq/criterions/__init__.py
urielsinger: Right, much clear :)

I added in the metaseq-inernal.{task/model}.__init__ references to metaseq's TASK_REGISTRY and MODEL_REGISTRY

PR #264, opened 2022-07-27
distributed main returns value up call stack

KUNAL1612:
**Patch Description**
Small change to ensure that distributed main returns value up the stack

**Testing steps**


suchenzang (approved):
Curious - how did this come up to add?

KUNAL1612:
> Curious - how did this come up as something to add?

So in #240 , I needed my main function to return a list of times taken by each thread to run generation. However by default no value was returned so I had to save it to an external file and read from it later. This was additional I/O cost as Stephen pointed out so returning the value up the call stack was an easier alternative, albeit it requiring this fix.

PR #265, opened 2022-07-27
fixing opt loading

lilisierrayu:
**Patch Description**
PR(https://github.com/facebookresearch/metaseq/pull/254) broke loading of old OPT model, this commit fixs it. 

**Testing steps**


suchenzang (approved):
Ahh thanks for catching! We should start thinking about how to add more testing around this "checkpoint loading" area of code, which seems to be rather brittle (I've broken this codepath before too). Opened https://github.com/facebookresearch/metaseq/issues/268 to track.

PR #267, opened 2022-07-28
updated to ensure cat happens on CUDA

KUNAL1612:
**Patch Description**
Observed during evaluating some models that the torch.cat operation in L732 took up 11.98GB of CPU memory. Used debugger to confirm that this was indeed because all tensors in shards_to_load function was on CPU. Because of this CPU operation, model loading using this function was slower than it needed to be. This speeds that up. 

**Testing steps**
Observed memory traces generated using profiler. Made changes and saw that model loading time using this function dropped anywhere between 10-15% given the same baseline conditions.

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -733,7 +733,7 @@ def _merge_flat_fsdp_shards(shards_to_load: List[Dict], unpad=False) -> Dict:
         dtype = shards_to_load[0]["model"][k].dtype
         if "flat_param" in k:
             pad_info_k = pad_info[k]
-            catted = torch.cat([x["model"][k] for x in shards_to_load])
+            catted = torch.cat([x["model"][k].cuda() for x in shards_to_load])
             if world_size == 1 and pad_info_k > 0:
                 catted = catted[:-pad_info_k]
             elif world_size > 1 and pad_info_k > 0 and not unpad:


suchenzang (approved):

KUNAL1612:
image

image


Before vs after on the operation after moving it to GPU

ngoyal2707:
wait wouldn't this then just take gpu memory and has potential to gpu oom?

ngoyal2707:
I think we load model states on cpu and build model on gpu and then do all the state loading.

This way would keep two copies on the gpu of the shard of weights.

KUNAL1612:
> wait wouldn't this then just take gpu memory and has potential to gpu oom?

I did not see overall increase in GPU memory . This is most likely because this tensor is eventually moved to GPU anyway.

KUNAL1612:
> I think we load model states on cpu and build model on gpu and then do all the state loading.
> 
> This way would keep two copies on the gpu of the shard of weights.

Interesting. Why are we loading model states on CPU then? I don't see the reason behind this at all unless we have a genuine risk of OOM issues (which also doesn't seem to be case)

ngoyal2707:
> > I think we load model states on cpu and build model on gpu and then do all the state loading.
> > This way would keep two copies on the gpu of the shard of weights.
> 
> Interesting. Why are we loading model states on CPU then? I don't see the reason behind this at all unless we have a genuine risk of OOM issues (which also doesn't seem to be case)

Cause then you will have two copies on gpu of the model shard that you are loading. This will quite likely gpu OOM the model hosting of 175B model on 8x80GB A100s

suchenzang:
@ngoyal2707 would it be too messy to try and gate this logic as a function of model size and available memory?

suchenzang (changes requested):
Deferring to @ngoyal2707

PR #273, opened 2022-07-29
hf compatibility test returns value + cleanup

KUNAL1612:
**Patch Description**
Owing to the changes in #264, there is now no longer a need to save the trimmed logits to an external file instead we could just return it directly. Also cleaned up a few unnecessary lines of code.


punitkoura (approved):
Thanks for cleaning this up!

PR #284, opened 2022-08-02
Validation accuracy during training

ramakanth-pasunuru:
- Tracks and plots validation accuracy for 2-way classification tasks during training
- Remove code owners for the opt_instruct branch

### DIFF
--- a/metaseq/model_parallel/criterions/vocab_parallel_cross_entropy.py
+++ b/metaseq/model_parallel/criterions/vocab_parallel_cross_entropy.py
@@ -9,7 +9,6 @@ import torch
 from metaseq import metrics, utils
 from metaseq.criterions import BaseCriterion, register_criterion
 
-
 try:
     from megatron.mpu.cross_entropy import (
         vocab_parallel_cross_entropy,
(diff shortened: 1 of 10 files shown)


metaseq/model_parallel/criterions/vocab_parallel_cross_entropy.py
```diff
@@ -45,6 +44,7 @@ def forward(self, model, sample, reduce=True):
         loss = vocab_parallel_cross_entropy(net_output[0].float(), target)
         if has_pad:
             loss = loss * (target != self.padding_idx)
+        batch_loss = loss.mean(-1)
```
stephenroller, line 47: I don't think we want this. We're going to be dividing by n twice now I believe?

stephenroller (changes requested):
Sorry Red flag

sriniiyer:
Did you do a before-after test? The train ppls are exactly the same before and after this PR, across muiltiple epochs?

ramakanth-pasunuru:
@sriniiyer Thanks for the suggestions. 
1. I checked that the train ppls remain the same for the whole training with and without this PR. 
2. I was able to match the reported accuracy during the training for valid set to be very close to the accuracy I got from gpt3_eval.py --> 47.4474 vs. 47.7477

sriniiyer (approved):
LGTM - This is awesome!

Issue #288, opened 2022-08-03
continue opt175b download from the latest available file


Barqawiz:
## 🚀 Feature Request
In the case of an internet issue or interruption, the download will stop, and the available opt175b download script will download from scratch.

### Motivation

opt175b files are enormous, and we need a way to continue the download from the last milestone. After two days of downloading the opt175b files, my two years old daughter unplugged the internet! And when I tried to re-run the download script, it started from scratch.

### Pitch

I create a script that downloads the file only if it does not exist or the file size is zero bytes and the script is ready to merge. 

### Alternatives

I checked the script files and I did not find a script to continue opt175b from latest milestone

### Additional context

The solution is available below as a new file if this feature is approved:
https://github.com/Barqawiz/metaseq/blob/main/metaseq/scripts/download_checkpoint_opt175b.sh

related pull request: #289

Barqawiz:
Merged #289 into main.

PR #289, opened 2022-08-03
continue opt175b download from the latest available file

Barqawiz:
**Patch Description**
Add script to continue the download after the latest available opt175b file in the case of an internet issue or interruption.

**Testing steps**
The following cases are tested:
- file does not exist.
- file exist with zero byte in case of running the script without internet.

related request #288


metaseq/scripts/download_checkpoint_opt175b.sh
```diff
@@ -0,0 +1,27 @@
...
+  do
+      filename="checkpoint_last-model_part-$part_id-shard$shard_id.pt"
+
+      if [ ! -s $filename ]; then
+        wget -O "$2$filename" "${presigned_url/$str_to_replace/$filename}"
+      else
+        echo "- ${filename} exist"
+      fi
+  done
+done
```
stephenroller, line 27:
```suggestion
done

```

stephenroller (commented):
leaving this one for @suchenzang to review but lgtm

suchenzang:
What's the reason for committing a new script instead of modifying metaseq/scripts/download_opt175b.sh ?

Barqawiz:
@suchenzang, to keep the choice of downloading everything from scratch if needed or using the new script to download from the last point. For any reason, if you want to merge it? I can do it. Otherwise, we can keep both options.

suchenzang:
Got it - yes, it would be cleaner to control through a flag from one file. And update documentation accordingly to highlight the two use-cases / ways to invoke the script (download from scratch vs from last saved).

Barqawiz:
ok sure will do it.

Barqawiz:
There are two parameters at the moment $1 mandatory for presigned_url and $2 optional for the directory. 

The available options to add a third optional parameter:
- **Option1**: keep the mandatory parameter $1 and add flags for both optional parameters regardless of the order. (-d for directory -c for checkpoint).
- **Option2**: move the directory parameter to $3 and use $2 for the checkpoint flag. 
- **Option3**: use $3 for the checkpoint flag and keep $2 for the directory. The issue of this option, the directory, should always be mentioned before the flag.

@suchenzang kindly advise which option is more suitable to start the implementation.

suchenzang:
@Barqawiz Option 1! Thanks for adding flags - it's definitely better than positional args :)

Barqawiz:
@suchenzang the merge is ready for your review.

The expected format, -c means continue download:
bash ./download_opt175b.sh "<presigned_url_given_in_email>" -d "<target directory>" -c

Also I added a validation if the user did not enter the presigned_url.

suchenzang (approved):
Amazing, thank you!!

Issue #285, opened 2022-08-02
_future_mask is created on CPU


KUNAL1612:
## 🐛 Bug

Using profiler to look for CPU intensive operations lead me to observe that _future_mask in Line [725](https://github.com/facebookresearch/metaseq/blob/c9c817d2a230519c2865264bafdf45931afa02e6/metaseq/models/transformer.py#L725) in transformer.py is being instantiated on GPU. While this is eventually being moved to GPU in [L747](https://github.com/facebookresearch/metaseq/blob/c9c817d2a230519c2865264bafdf45931afa02e6/metaseq/models/transformer.py#L747), the operations that happen between L728 to L745 happen on CPU since the tensor itself is on CPU. 

However the code follows that logic only during document attention when the doc separator is not -1. So in effect, creating self._future_mask on GPU in L725 would only offer marginal improvements in some very few specific scenarios. 

Need advice on if this change is worth making

sriniiyer:
Adding @ramakanth-pasunuru and @punitkoura, who have looked at this before

punitkoura:
This code is used for implementing document attention... We were using it for finetuning, but not sure if it is still being used. @ramakanth-pasunuru / @sriniiyer might know

KUNAL1612:
My pitch is that because we move the _future_mask to GPU in line [747](https://github.com/facebookresearch/metaseq/blob/c9c817d2a230519c2865264bafdf45931afa02e6/metaseq/models/transformer.py#L747), we don't incur any extra memory or processing overhead by moving it before so that if at some point we use the code path we do get the speedup

sriniiyer:
We might use this very soon, so this might be worth doing. @ramakanth-pasunuru

ramakanth-pasunuru:
@KUNAL1612 sounds good to move the future mask to GPU. Document attention is useful for finetuning and I will run some experiments in the coming weeks on that.

PR #290, opened 2022-08-04
moving future_mask to cuda for document attenion

KUNAL1612:
**Patch Description**
Moved the future mask to CUDA so that all operations for document attention also take place on CUDA. Refer to issue #285 for context. The speedup offered by this is marginal but over multiple runs may cumulate.

To test, I set attn doc seperator to a random number to trigger the code to enter the branch, and timed it using the methods used in #220. Observed that over 436 calls to the function on CPU it takes 0.42 seconds vs 0.09 seconds on GPU.

### DIFF
--- a/metaseq/models/transformer.py
+++ b/metaseq/models/transformer.py
@@ -724,7 +724,7 @@ class TransformerDecoder(IncrementalDecoder):
         if need_to_make_new_mask:
             self._future_mask = torch.triu(
                 utils.fill_with_neg_inf(torch.zeros([max_seq_len, max_seq_len])), 1
-            )
+            ).to(tensor)
             if self.self_attn_doc_sep != UNSPECIFIED_DOC_SEP:
                 # Code to accomodate dynamic attention when document seperator is used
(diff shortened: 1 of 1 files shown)


punitkoura (approved):

metaseq/models/transformer.py
```diff
@@ -724,7 +724,7 @@ def buffered_future_mask(self, tensor, input_tokens=None):
         if need_to_make_new_mask:
             self._future_mask = torch.triu(
                 utils.fill_with_neg_inf(torch.zeros([max_seq_len, max_seq_len])), 1
-            )
+            ).to(tensor)
```
ngoyal2707, line 727: instead of creating the tensor on cpu and moving to cuda,

you could just create on cuda like:
```
self._future_mask = torch.triu(
        utils.fill_with_neg_inf(torch.zeros([max_seq_len, max_seq_len], device=tensor.device)), 1
)
```

KUNAL1612: updated to address this

PR #297, opened 2022-08-09
Fix the batch size for 175b benchmark

jieru-hu:
**Patch Description**
Fix the batch size for 175b benchmark per offline discussions with Stephen

**Testing steps**
Run the benchmark on Azure.


stephenroller (approved):
Thank you!

PR #303, opened 2022-08-11
Ensuring that the API returns None as the first probability in echo mode

punitkoura:
**Patch Description**
Describe your changes

**Testing steps**

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -678,9 +678,15 @@ class GeneratorInterface:
                                 : max_tokens[i]
                             ]
 
-                    # cut off the starting token
+                    # cut off the starting </s> token and the first token, and attach
+                    # a None at the beginning.
+                    # So, if the tokens are </s>, The, nice, cat
+                    # scores are            -0.5, -0.3, -0.2, -0.1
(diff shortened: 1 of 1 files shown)


metaseq/hub_utils.py
```diff
@@ -678,21 +678,29 @@ def generate(
...
+                    # So, if the tokens are </s>, The, nice, cat
+                    # scores are            -0.5, -0.3, -0.2, -0.1
+                    # We would remove the scores for </s> and the first word "The"
+                    # and attach a None to signify the score for "The"
+                    # This is in line with how the OpenAI API operates.
+                    assert tokens[0] == self.task.target_dictionary.eos_index
+
                     tokens_no_eos = tokens[1:] if echo else tokens
-                    scores_with_eos = [None] + scores[1:] if echo else scores
+                    scores_no_eos = [None] + scores[2:] if echo else scores
```
stephenroller, line 691: eeee i'm not sure if this is the implementation we want

lilisierrayu: I am with Stephen here. I prefer the model return scores[1:] here.

PR #305, opened 2022-08-16
[jsonl] Support other tokenizers.

stephenroller:
**Patch Description**
Add support for other HF tokenizers for when we train our own.

**Testing steps**
Started training OPT baseline with new tokenizer.

### DIFF
--- a/metaseq/tasks/streaming_language_modeling.py
+++ b/metaseq/tasks/streaming_language_modeling.py
@@ -31,7 +31,7 @@ from metaseq.dataclass import MetaseqDataclass
 from metaseq.tasks import LegacyTask, register_task
 
 try:
-    from tokenizers import ByteLevelBPETokenizer
+    from tokenizers import ByteLevelBPETokenizer, Tokenizer
 
     has_hf_tokenizers = True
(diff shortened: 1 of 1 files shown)


metaseq/tasks/streaming_language_modeling.py
```diff
@@ -124,9 +127,12 @@ def __init__(self, args):
         if not has_hf_tokenizers:
             raise ImportError("Please install tokenizers with: pip install tokenizers")
 
-        self.tokenizer = ByteLevelBPETokenizer.from_file(
-            args.vocab_filename, args.merges_filename
-        )
+        if args.hf_tokenizer:
```
suchenzang, line 130: Is there no way to unify the two somehow?  AKA rewrite the vocab/merges filename into the expected `Tokenizer.from_file(...)` input?

stephenroller: Yes, I would like that but I can't afford to do the resulting refactor in other places (e.g. the API) right now.

#308 to mark it.

suchenzang (approved):

PR #312, opened 2022-08-20
[jsonl] Fix regression in index building.

stephenroller:
**Patch Description**
There was a performance regression in the JSONL dataset indexer, where when building the index it parsed the jsonl files when it really just needed start-of-line numbers. This was originally done to skip over bad parses.

This PR returns us to the much older behavior of throwing a hard crash, midway during training, when we come to a bad data point. On the positive side, we still report _where_ the file was bad.

We may need to build a separate validator to deal with bad data parses, but at the end of the day, it doesn't feel like "skipping bad data" is the right thing to do.

**Testing steps**
In progress

### DIFF
--- a/metaseq/data/jsonl_dataset.py
+++ b/metaseq/data/jsonl_dataset.py
@@ -91,9 +91,20 @@ class JsonlDataset(torch.utils.data.Dataset):
         if subshard_idx < 0 or subshard_idx >= len(self.offsets):
             raise IndexError
         f = self._get_mmap()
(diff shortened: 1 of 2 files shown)


suchenzang (approved):

metaseq/data/jsonl_dataset.py
```diff
@@ -128,15 +139,6 @@ def _build_index(self, path: str):
         line_num = 0
```
suchenzang, line 139: Nit: `line_num` seems extraneous now

PR #317, opened 2022-08-23
added code to disable bias

ngoyal2707:
Gives 2-3% speed up with almost no ppl loss, plus it is very hard to have correct bias gradients with sequence parallel + FSDP for now


suchenzang (approved):
Nit: lint test :)

PR #322, opened 2022-08-30
[checkpoints] Don't validate/save after epochs.

stephenroller:
**Patch Description**
Our current logic always validates and saves at the end of an epoch, even when we set flags such that it isn't supposed to. This is because we ignored the "no_epoch_checkpoints" flag, and because the "validate_interval" flag made it impossible (as it's an interval of epochs). I've killed the validate_interval flag and updated the logic appropriately.

**Testing steps**
Been training with this patch.

### DIFF
--- a/metaseq_cli/train.py
+++ b/metaseq_cli/train.py
@@ -402,7 +402,11 @@ def validate_and_save(
         )
 
     do_save = (
-        (end_of_epoch and epoch_itr.epoch % cfg.checkpoint.save_interval == 0)
+        (
+            end_of_epoch
+            and epoch_itr.epoch % cfg.checkpoint.save_interval == 0
+            and not cfg.checkpoint.no_epoch_checkpoints
+        )
         or should_stop
         or (
             cfg.checkpoint.save_interval_updates > 0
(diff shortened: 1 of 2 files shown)


metaseq_cli/train.py
```diff
@@ -402,7 +402,11 @@ def validate_and_save(
         )
 
     do_save = (
-        (end_of_epoch and epoch_itr.epoch % cfg.checkpoint.save_interval == 0)
+        (
+            end_of_epoch
+            and epoch_itr.epoch % cfg.checkpoint.save_interval == 0
+            and not cfg.checkpoint.no_epoch_checkpoints
```
suchenzang, line 408: Hm... this flag is still a bit confusing - likely since there's just way too many no_* checkpointing related flags (https://github.com/facebookresearch/metaseq/blob/414c32dfbc0789421aeceb6d979fb0f214e82dc1/metaseq/dataclass/configs.py#L514-L530). Any chance we can clean all of these up?

stephenroller: You can clean it up if you want but maybe the thing to do is change the help msg for no_epoch_checkpoints is "don't store checkpoints at epoch boundaries"

suchenzang: Yeah let's start with that help message change for now for this PR. I'll open an issue to track the above cleanup.

stephenroller: Give me the conditional approve then?

suchenzang (approved):

PR #330, opened 2022-09-06
BF16 bug fix:

urielsinger:
### **in non FSDP mode**
when not in FSDP, the model wasn't converted to bf16 although bf16 was set to True.

### **optimizer._multiply_factor reset to 1.0 each step**
In MemoryEfficientFP16Optimizer.zero_grad (that is applied every training step), the model sets  _multiply_factor back to 1.0:
https://github.com/facebookresearch/metaseq/blob/bbcedfebb4c35f71cdda1f1a358491f3996a9fc3/metaseq/optim/fp16_optimizer.py#L452
A similar thing was applied also in the normal FP16Optimizer.

### DIFF
--- a/metaseq/optim/fp16_optimizer.py
+++ b/metaseq/optim/fp16_optimizer.py
@@ -241,7 +241,8 @@ class _FP16OptimizerMixin(object):
 
         if self.scaler is not None:
             self._multiply_factor = 1.0 / float(self.scaler.loss_scale)
-
+        else:
+            self._multiply_factor = 1.0
 
 class FP16Optimizer(_FP16OptimizerMixin, optim.BaseOptimizer):
     """
--- a/metaseq/optim/fp16_optimizer.py
+++ b/metaseq/optim/fp16_optimizer.py
@@ -244,6 +244,7 @@ class _FP16OptimizerMixin(object):
         else:
             self._multiply_factor = 1.0
 
+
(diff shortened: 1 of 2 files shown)


stephenroller (approved):

urielsinger:
@stephenroller 
Just out of curiosity, this _multiply_factor bug is a must in-order to train with BF16 (otherwise, it crashes to zero after a while), so how have you manage to train with BF16?

ngoyal2707:
> @stephenroller Just out of curiosity, this _multiply_factor bug is a must in-order to train with BF16 (otherwise, it crashes to zero after a while), so how have you manage to train with BF16?

When training with DDP in mixed prevision mode, the model code creates a bf16 / fp16 model copy and the `fp16_optimizer` code creates and keeps another fp32 copy.

When training with FSDP in mixed precision mode, the model code creates a fp32 model copy and FSDP code internally keeps a fp16 / bf16 model copy. 
So with FSDP, we always use "MemoryEfficientOptimizer" class with or without mixed precision cause the optimizer is not maintaining the extra set of model weights (logic here: https://github.com/facebookresearch/metaseq/blob/main/metaseq/trainer.py#L245)
and the scale factor is correctly set for that part of code path: https://github.com/facebookresearch/metaseq/blob/main/metaseq/optim/fp16_optimizer.py#L449

PR #331, opened 2022-09-06
[dist] Force threads=1 when spawning.

stephenroller:
**Patch Description**
Some of our consolidation logic is extremely slow: this is due to the concat's happening on CPU, and the default in pytorch is to use OMP threads to parallelize this. However, we already have spawned multiple processes (or slurm did), so all the threads end up fighting and thrashing for the CPUs.

We could do this individually in every script if we wanted to be careful about optimizing for CPU cases, but given that we basically only ever do GPU logic, we can just bake this assumption right into our distributed code.

**Testing steps**
Before change, loading checkpoints pegged all CPUs to 100%. After change, they do not and it runs faster.

### DIFF
--- a/metaseq/distributed/utils.py
+++ b/metaseq/distributed/utils.py
@@ -187,6 +187,8 @@ def distributed_init(cfg: MetaseqConfig):
 
 
 def distributed_main(i, main, cfg: MetaseqConfig, kwargs):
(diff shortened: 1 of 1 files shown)


stephenroller:
Probably not in specific circumstances.

PR #333, opened 2022-09-07
alibi performance fixes

ruanslv:
Attempt to reduce memory usage from alibi embeddings. Running some ablations to test

### DIFF
--- a/metaseq/models/transformer.py
+++ b/metaseq/models/transformer.py
@@ -741,17 +741,25 @@ class TransformerDecoder(IncrementalDecoder):
                     ] = float("-inf")
 
             if self.use_alibi:
+                alibi = self.alibi.to(tensor)
+                print("org")
+                print(self._future_mask.shape)
                 self._future_mask = (
                     self._future_mask.unsqueeze(0)
                     .expand(self.args.decoder_attention_heads, max_seq_len, max_seq_len)
                     .contiguous()
-                    + self.alibi
+                    + alibi
                 )
+                print("alibi")
+                print(self._future_mask.shape)
 
         self._future_mask = self._future_mask.to(tensor)
         if self.use_alibi:
             future_mask = self._future_mask[:, :cur_seq_len, :cur_seq_len]
-            return future_mask.repeat(batch_size, 1, 1)
+            future_mask = future_mask.repeat(batch_size, 1, 1)
+            print("final")
+            print(self._future_mask.shape)
+            return future_mask
         elif self.self_attn_doc_sep != UNSPECIFIED_DOC_SEP:
             return self._future_mask
         else:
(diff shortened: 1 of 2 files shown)


metaseq/models/transformer.py
```diff
@@ -746,16 +741,17 @@ def buffered_future_mask(self, tensor, input_tokens=None):
                     ] = float("-inf")
 
             if self.use_alibi:
-                alibi = self.alibi.repeat(batch_size, 1, 1)  # batch_size, 1, 1
-                self._future_mask = self._future_mask.unsqueeze(0) + alibi
+                self._future_mask = (
+                    self._future_mask.unsqueeze(0)
+                    .expand(self.args.decoder_attention_heads, max_seq_len, max_seq_len)
+                    .contiguous()
```
stephenroller, line 747: i wonder why the contiguous is necessary

let's make sure to benchmark speed too!

stephenroller:
@ofirpress would you mind reviewing too?

ofirpress:
LGTM, but the way to fully verify this is to train a model with it and compare it to a sinusoidal or learned position emb baseline.

You should see:
1. ALiBi should run just as fast as the baseline during train/inf.
2. ALiBi should use at most ~200MB of extra memory, not more. 
3. ALiBi PPL should be slightly better (even when we train and evaluate on sequences of the same length).
4. ALiBi should be able to handle sequences that are 50% larger and PPL should go down. (So for example if you train on 1000 tokens and get 12 PPL at inference with 1000 tokens, try evaluating on 1500 tokens and you should see PPL go down, maybe to like 11...)

If this doesn't occur then ping me and me and Mike will look into it

suchenzang (approved):

ruanslv:
A couple of issues I found so far to integrate Alibi with our code:

(Minor / fixable):
- One host to GPU transfer required, alibi bias tensor was in CPU;
- We compute multihead attention using model parallelism split across attention heads, and given alibi generates different mask values per head, we have to split the mask across the same dimension too.

(Maybe serious problem):
- We rely on megatron fused kernels to apply mask + compute softmax, and they only work when the mask across attention heads is the same. I took a brief look at the megatron kernel code but too obscure for me, couldn't figure out how to generalize it: https://github.com/NVIDIA/Megatron-LM/blob/e156d2fea7fc5c98e645f7742eb86b643956d840/megatron/fused_kernels/scaled_masked_softmax_cuda.cu#L43

I don't know however how important just this kernel is to us. Megatron paper claimed about ~10% throughput increase across all fused kernels they used, but IIUC that included fused kernels for element-wise operations too (bias +
GeLU and bias + dropout + add): https://arxiv.org/pdf/2104.04473.pdf

cc @ngoyal2707 in case you have any thoughts

ofirpress:
Some thoughts:
> We compute multihead attention using model parallelism split across attention heads, and given alibi generates different mask values per head, we have to split the mask across the same dimension too.

Please tell me if this becomes too much of an issue. Me and Stephen talked about how if you have 96 heads lets say, its totally fine to break them up into groups of 8 or 16 and have each group use the same ALiBi slope, if that makes things faster or easier. 
But I also think it would be possible to make this work without that workaround. 

> megatron
I know very little about this, so I might be totally wrong here, but BigScience used megatron, maybe they've solved this problem?

stephenroller:
Abandoning for concerns with integration with flashattention.

PR #335, opened 2022-09-09
multiclass val accuracy during finetuning

ramakanth-pasunuru:
Changes:
- remove sampling for validation splits
- update validation accuracy to support multi-label (previously we support only two-label tasks) during finetuning

### DIFF
--- a/metaseq/data/streaming_src_tgt_dataset.py
+++ b/metaseq/data/streaming_src_tgt_dataset.py
@@ -123,9 +123,9 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
     cur_tgt_block = []
     cur_block_remain = block_size
     for idx, (src, tgt) in enumerate(iterable):
-        if isinstance(
-            src, tuple
-        ):  # Adding negative examples not supported in this mode
+        if isinstance(src, tuple):
+            # When we get a tuple it means that we are also processing negative candidates
+            # Adding negative examples not supported in this mode
             src = src[0]
             tgt = tgt[0]
         if src.numel() > block_size:
@@ -185,6 +185,8 @@ def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, paddin
     no_neg_examples = False
     for all_src, all_tgt in iterable:
         if not isinstance(all_src, tuple):
+            # When its not a tuple, example doesn't come with negative candidates
+            # So yield the blocks accordingly.
             no_neg_examples = True
             all_src = (all_src,)
             all_tgt = (all_tgt,)
--- a/metaseq/data/streaming_src_tgt_dataset.py
+++ b/metaseq/data/streaming_src_tgt_dataset.py
@@ -53,8 +53,6 @@ class StreamingSrcTgtDataset(torch.utils.data.IterableDataset):
             self.block_iterator = yield_src_tgt_blocks
         elif break_mode == "eos_pad_8":  # Single example per sequence
             self.block_iterator = yield_src_tgt_single_sentences_pad_8
-        elif break_mode == "eos_neg_pad_8":
-            self.block_iterator = yield_src_tgt_single_sentences_withneg_pad_8
         else:
             raise NotImplementedError(f"Unknown break mode: {break_mode}")
 
@@ -125,7 +123,9 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
     cur_tgt_block = []
     cur_block_remain = block_size
     for idx, (src, tgt) in enumerate(iterable):
-        if isinstance(src, tuple):
+        if isinstance(
+            src, tuple
+        ):  # Adding negative examples not supported in this mode
             src = src[0]
             tgt = tgt[0]
         if src.numel() > block_size:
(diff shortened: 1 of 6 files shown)


metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -207,3 +215,49 @@ def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, paddin
...
+        if not isinstance(all_src, tuple):
+            all_src = (all_src,)
+            all_tgt = (all_tgt,)
+        for idx, (src, tgt) in enumerate(zip(all_src, all_tgt)):
+            cur_src_block = []
+            cur_src_block_ids = []
+            cur_tgt_block = []
+            if src.numel() > block_size:
+                # truncate right side
+                # TODO: Enable left side truncation
```
todpole3, line 237: Switch to left-truncation now?

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -207,3 +215,49 @@ def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, paddin
...
+            cur_block_remain -= src.numel()
+            padding = cur_src_block[-1].new_full((cur_block_remain,), padding_idx)
+            cur_src_block.append(padding)
+            cur_tgt_block.append(padding)
+
+            yield {
+                "ids": torch.LongTensor(cur_src_block_ids),
+                "src_block": torch.cat(cur_src_block),
+                "tgt_block": torch.cat(cur_tgt_block),
+                "is_positive": torch.tensor([1 if idx == 0 else 0]).bool(),
```
todpole3, line 261: This function is a duplicate of yield_src_tgt_single_sentences_pad_8 with two additional fields in each returned element, right? Perhaps be more explicit about this in the description in line 224. Alternatively, can this be implemented in yield_src_tgt_single_sentences_pad_8 with a boolean argument to control whether to add the two additional fields?

metaseq/model_parallel/criterions/vocab_parallel_cross_entropy.py
```diff
@@ -44,7 +44,7 @@ def forward(self, model, sample, reduce=True):
         loss = vocab_parallel_cross_entropy(net_output[0].float(), target)
         if has_pad:
             loss = loss * (target != self.padding_idx)
-        batch_loss = loss.mean(-1)
+        batch_loss = loss.sum(-1)  # TODO: length normalize?
```
todpole3, line 47: This seems to be an important change, what's the motivation?

metaseq/tasks/streaming_finetune_language_modeling.py
```diff
@@ -106,7 +125,7 @@ def load_dataset(self, split: str, epoch=1, combine=False, **kwargs):
             block_size=self.args.tokens_per_sample + 1,
             break_mode=self.args.sample_break_mode
             if split == "train"
-            else "eos_pad_8",  # eos mode for val/test to get accuracies
+            else self.args.valid_sample_break_mode,  # use diferent mode for val/test
```
todpole3, line 128: What are the different modes for valid and test?

ramakanth-pasunuru:
Refactored code such that instead of using different break mode for calculating validation accuracy, data reading way is changed.

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -53,6 +53,8 @@ def __init__(
             self.block_iterator = yield_src_tgt_blocks
         elif break_mode == "eos_pad_8":  # Single example per sequence
             self.block_iterator = yield_src_tgt_single_sentences_pad_8
+        elif break_mode == "eos_neg_pad_8":
```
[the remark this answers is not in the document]
ramakanth-pasunuru: After your comment, I removed relying on different break mode, instead directly change the way we read data for calculating validation accuracy. See my new commits.

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -123,6 +125,9 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
     cur_tgt_block = []
     cur_block_remain = block_size
     for idx, (src, tgt) in enumerate(iterable):
+        if isinstance(src, tuple):
```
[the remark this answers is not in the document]
ramakanth-pasunuru: When they have negative candidates

metaseq/data/streaming_src_tgt_dataset.py
ramakanth-pasunuru: removed this module

metaseq/model_parallel/criterions/vocab_parallel_cross_entropy.py
ramakanth-pasunuru: mean was added just in my previous PR. We need to see what makes most sense.

metaseq/tasks/streaming_finetune_language_modeling.py
ramakanth-pasunuru: outdated. No using this anymore

metaseq/data/streaming_src_tgt_dataset.py
ramakanth-pasunuru: this module is removed, but it's a good point. Let's decide on this truncation method and update it in a separate PR.

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -123,6 +123,11 @@ def yield_src_tgt_blocks(iterable, block_size, drop_last, padding_idx):
     cur_tgt_block = []
     cur_block_remain = block_size
     for idx, (src, tgt) in enumerate(iterable):
+        if isinstance(
+            src, tuple
+        ):  # Adding negative examples not supported in this mode
```
stephenroller, line 128: I'm not sure what's going on here. Can we add some comments explaining why we might get tuples?


Nit: move the comment inside the if block

metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -176,34 +181,50 @@ def yield_src_tgt_single_sentences_pad_8(iterable, block_size, drop_last, paddin
     return the example as is, without packing, truncating to block_size in cases of
     very long examples.
     """
+    counter = 0
+    no_neg_examples = False
+    for all_src, all_tgt in iterable:
```
stephenroller, line 186: In general I feel like we're starting to have a lot of ambiguous types and wondering if we shouldn't subclass for your case, rather than trying to bring more logic into SLM?

stephenroller (commented):
Broadly this feels like a lot of project-oriented logic bubbling up into top level abstractions. Let's think about how we could refactor, perhaps by subclassing losses/tasks and providing additional task-specific hooks to call from train.py

sriniiyer (approved):
LGTM

metaseq/data/streaming_src_tgt_dataset.py
ramakanth-pasunuru: updated.

metaseq/data/streaming_src_tgt_dataset.py
ramakanth-pasunuru: Thanks for the suggestions. I think adding negative examples would be in general useful (if not now, later) incase if we implement different losses that consider negative examples as well. So, I feel like SFLM shud have this capability of processing negatives. Probably, the current implementation is not the best to add this feature, but we will improve it later before merging into main branch. Currently, moving on to unblock some experiments.

stephenroller:
Wait no. I didn't want this merged.

stephenroller:
Oh this wasn't merging into Main. Okay no worries.

PR #338, opened 2022-09-14
Remove no_ and keep_ checkpoint flags

Xirider:
**Patch Description**
Removed most of the no_* and keep_* flags and cleaned up their logic.

**Testing steps**
I did have some issues with setting up an environment to test this.

In the end I tested with the opt_baselines script and checked if different `save_intervals `and `save_interval_updates ` correctly create all the checkpoints:
`python3 metaseq/launcher/opt_baselines.py --prefix train.8m --model-size 8m --checkpoints-dir ./test-checkpoint --tensorboard-logdir ./test-checkpoint --num-trials 1 --azure --num-gpus 4 --num-nodes 1 --seed 1 --circleci --local --disable-validation --max-epoch 5 --max-update 200`

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -68,7 +68,9 @@ def save_checkpoint(
     checkpoint_conds = collections.OrderedDict()
 
     save_for_epoch = (
-        end_of_epoch and cfg.save_interval > 0 and epoch % cfg.save_interval == 0
+        end_of_epoch
+        and cfg.save_interval_epochs > 0
+        and epoch % cfg.save_interval_epochs == 0
     )
 
     save_for_updates = (
@@ -88,34 +90,64 @@ def save_checkpoint(
     checkpoints = [
         os.path.join(cfg.save_dir, fn) for fn, cond in checkpoint_conds.items() if cond
     ]
+    if len(checkpoints) > 0:
+        if PathManager.islink(checkpoints[0]):
+            PathManager.rm(checkpoints[0])
 
-    if PathManager.islink(checkpoints[0]):
-        PathManager.rm(checkpoints[0])
-
-    trainer.save_checkpoint(
-        checkpoints[0],
-        extra_state,
-        training_finished=training_finished,
-        async_callback_fn=async_callback_fn,
-    )
-
-    def _copy_if_not_async(src, dest):
-        if cfg.write_checkpoints_asynchronously:
-            pass  # TODO[ioPath]: Need to implement a delayed asynchronous file copying/moving feature.
-        else:
-            assert PathManager.copy(
-                src, dest, overwrite=True
-            ), f"Failed to copy {src} to {dest}"
-
-    for cp in checkpoints[1:]:
-        _copy_if_not_async(src=checkpoints[0], dest=cp)
-
-    write_timer.stop()
-    logger.info(
-        "Saved checkpoint {} (epoch {} @ {} updates, score {}) (writing took {} seconds)".format(
-            checkpoints[0], epoch, updates, val_loss, write_timer.sum
+        trainer.save_checkpoint(
+            checkpoints[0],
+            extra_state,
+            training_finished=training_finished,
+            async_callback_fn=async_callback_fn,
         )
-    )
+
+        def _copy_if_not_async(src, dest):
+            if cfg.write_checkpoints_asynchronously:
+                pass  # TODO[ioPath]: Need to implement a delayed asynchronous file copying/moving feature.
+            else:
+                assert PathManager.copy(
+                    src, dest, overwrite=True
+                ), f"Failed to copy {src} to {dest}"
+
+        for cp in checkpoints[1:]:
+            _copy_if_not_async(src=checkpoints[0], dest=cp)
+
+        write_timer.stop()
+        logger.info(
+            "Saved checkpoint {} (epoch {} @ {} updates, score {}) (writing took {} seconds)".format(
+                checkpoints[0], epoch, updates, val_loss, write_timer.sum
+            )
+        )
+
+        _delete_old_checkpoint_files(
+            cfg,
+            end_of_epoch,
+            suffix,
+        )
+
+
+def _delete_old_checkpoint_files(
+    cfg: CheckpointConfig, end_of_epoch: bool, suffix: str
+):
+    if not end_of_epoch and cfg.keep_last_updates > 0:
+        suffixes = [suffix]
+
+        # remove old checkpoints; checkpoints are sorted in descending order
+        for one_suffix in suffixes:
+            checkpoints = _checkpoint_paths(
+                cfg.save_dir, pattern=r"checkpoint_(\d+){}\.pt".format(one_suffix)
+            )
+            for old_chk in checkpoints[cfg.keep_last_updates :]:
+                if os.path.lexists(old_chk):
+                    os.remove(old_chk)
+    if cfg.keep_last_epochs > 0:
+        # remove old epoch checkpoints; checkpoints are sorted in descending order
+        checkpoints = _checkpoint_paths(
+            cfg.save_dir, pattern=r"checkpoint(\d+){}\.pt".format(suffix)
+        )
+        for old_chk in checkpoints[cfg.keep_last_epochs :]:
+            if os.path.lexists(old_chk):
+                os.remove(old_chk)
 
 
 def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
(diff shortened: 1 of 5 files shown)


metaseq/checkpoint_utils.py
```diff
@@ -105,77 +84,34 @@ def is_better(a, b):
     checkpoints = [
         os.path.join(cfg.save_dir, fn) for fn, cond in checkpoint_conds.items() if cond
     ]
-    if len(checkpoints) > 0:
-        if PathManager.islink(checkpoints[0]):
-            PathManager.rm(checkpoints[0])
```
suchenzang, line 110: Hm.. I think we might still want to keep this cleanup logic - what was the reason for removal?

metaseq/trainer.py
```diff
@@ -325,8 +323,7 @@ def consolidate_optimizer(self):
     def state_dict(self, filename, training_finished=False) -> Dict[str, Dict]:
         model_state_dict = self.model.state_dict()
         optim_state = None
```
suchenzang, line 325: Can remove this now

metaseq/checkpoint_utils.py
```diff
@@ -47,11 +47,7 @@ def save_checkpoint(
         best_function = max if cfg.maximize_best_checkpoint_metric else min
         save_checkpoint.best = best_function(val_loss, prev_best)
 
-    if cfg.no_save:
```
stephenroller, line 50: --no-save is still useful for benchmarking. To this end, the warning is still helpful.

metaseq/checkpoint_utils.py
```diff
@@ -71,32 +67,15 @@ def is_better(a, b):
     suffix = trainer.checkpoint_suffix
     checkpoint_conds = collections.OrderedDict()
     checkpoint_conds[f"checkpoint{epoch}{suffix}.pt"] = (
-        end_of_epoch and not cfg.no_epoch_checkpoints and epoch % cfg.save_interval == 0
```
stephenroller, line 74: i am early in review but now this reads like we will ALWAYS checkpoint at end of epoch.

Which is not the worst (it's actually kinda convenient for me) but hmmm. Want discussion from others on the team.

metaseq/checkpoint_utils.py
```diff
@@ -71,32 +67,15 @@ def is_better(a, b):
...
-        and not cfg.no_best_checkpoints
-    ):
-        checkpoint_conds[
-            "checkpoint.best_{}_{:.2f}.pt".format(cfg.best_checkpoint_metric, val_loss)
-        ] = not hasattr(save_checkpoint, "best") or is_better(
-            val_loss, save_checkpoint.best
-        )
-    checkpoint_conds[f"checkpoint_last{suffix}.pt"] = not cfg.no_last_checkpoints
+
+    checkpoint_conds[f"checkpoint_last{suffix}.pt"] = True
```
stephenroller, line 78: commenting here on line 78 for reference

metaseq/checkpoint_utils.py
stephenroller: looks like w're mostly getting rid of the "if len(checkpoints) > 0" line and unindenting the rest. This seems congruent with L78 hardcoding that the size of checkpoint_conds will always be >= 1

stephenroller (commented):
awesome, thanks for working on this!

metaseq/checkpoint_utils.py
Xirider: Yes, Stephen is right. It is only removing  "if len(checkpoints) > 0" and unindenting the rest. Since we removed all no_save* flags, there is no way for "save_checkpoint" to be called without actually saving at least one checkpoint.

metaseq/checkpoint_utils.py
Xirider: So, should I add `--no-save` back in?

metaseq/checkpoint_utils.py
Xirider: It might make more sense to also check for `save_interval > 0`, similar to how it is done for `save_interval_updates`. This way you always have the option to turn of checkpointing by setting both `--save-interval` and `--save-interval-updates` to 0.

metaseq/checkpoint_utils.py
```diff
@@ -105,77 +84,34 @@ def is_better(a, b):
...
+        PathManager.rm(checkpoints[0])
 
-        write_timer.stop()
-        logger.info(
-            "Saved checkpoint {} (epoch {} @ {} updates, score {}) (writing took {} seconds)".format(
-                checkpoints[0], epoch, updates, val_loss, write_timer.sum
-            )
-        )
-
-    _delete_old_checkpoint_files(
```
suchenzang, line 137: cc @aitorormazabal given your latest PR: https://github.com/facebookresearch/metaseq/pull/341 - it seems like you are calling into this codepath; what kind of checkpoint deletion frequency/policy are you looking for?

metaseq/checkpoint_utils.py
suchenzang: Can we gate this to depend on the `--benchmark` flag then?

metaseq/checkpoint_utils.py
suchenzang: Hm.. can we remove any possibility of checkpointing outside of the frequency of checkpoint interval (purely as a function of # of iterations) just to simplify for the time-being?

metaseq/checkpoint_utils.py
aitorormazabal: Yes I am using `--keep-interval-updates 2 ` to prevent my checkpoint dirs from growing too much. I find the `keep_` functionality quite useful for running experiments with limited disk space

suchenzang: Thanks @aitorormazabal - sorry @Xirider we're now using your PR to gather requirements live 😅 . Seems like we should include a "keep last N checkpoints" logic to handle the above. 

Net diff in this PR will now give us only two knobs: checkpoint every N number of iterations, keep last K checkpoints (note that each checkpoint can be many files, so this "last K" logic should account for that; would be good to confirm since we haven't turned this on for large-scale training yet).

metaseq/checkpoint_utils.py
Xirider: I simplified further so that there is no possibility of checkpointing outside of the frequency of checkpoint interval.

metaseq/checkpoint_utils.py
Xirider: I changed benchmark in opt_baseline now to set `--save-interval` and `--save-interval-updates` to 0, which I felt is cleaner than adding an additional flag just for this use case. If you feel different, I can add a new flag back as well.

metaseq/checkpoint_utils.py
Xirider: 😂 @suchenzang So do we want still want to be able to save after epochs? And do we want to save after a run is finished early?

--save-interval
--save-interval-updates

--keep-interval-updates
--keep-interval

or just:

--save-interval-updates
--keep-interval-updates

suchenzang: Ah yeah forgot we're keeping those two (epoch vs interval) separate 🤦‍♀️ . So yes, the first case:
```
--save-interval
--save-interval-updates

--keep-interval
--keep-interval-updates
```
where the former is a frequency, and the latter is a count (hopefully that isn't confusing to distinguish between).

To walk through an example: I would imagine wanting to set save every epoch on + save every 100 steps, followed by keep last epoch (evict older ones) + last 5 checkpoints (mapping to 500 steps) so that we don't ever have more than 1+5 checkpoints around on disk (in the above example).

Xirider: Yes, that makes sense. Thanks for the example!

Xirider:
- renamed `--save-interval` to `--save-interval-epochs`, otherwise it will be quite confusing with the keep flags as well.

The final flags are now:
save_interval_epochs
save_interval_updates
keep_last_epochs
keep_last_updates

- tested multiple combinations of them, as well as --benchmark.

vesko-st:
It looks like this diff broke metaseq-internal scripts: metaseq_internal/fb_sweep/sweep_openlm_baselines.py

suchenzang:
Ah, some sweep scripts were updated (https://github.com/fairinternal/metaseq-internal/pull/431) but not all.  I just pinned the post in our internal workplace group notifying of the breaking change to sweep scripts - hopefully that can help alert of these changes in this future!

Xirider:
> It looks like this diff broke metaseq-internal scripts: metaseq_internal/fb_sweep/sweep_openlm_baselines.py

This is now fixed in https://github.com/fairinternal/metaseq-internal/pull/453

PR #339, opened 2022-09-15
add control option to turn on validation at step 0

todpole3:
**Patch Description**
Add option to run validation of checkpoint before any training or fine-tuning starts


ramakanth-pasunuru (approved):
LGTM

PR #341, opened 2022-09-16
Change deletion of old checkpoints to match new naming

aitorormazabal:
**Patch Description**
Fix old checkpoint deletion code to match new naming from https://github.com/facebookresearch/metaseq/commit/6c02d8311f3d7db994d1908edca6096eb99ff16d 
**Testing steps**

Checked old checkpoints are being deleted now.


suchenzang (approved):
Thanks for fixing!

PR #342, opened 2022-09-16
test removing output projection from MHA

ruanslv:
**Patch Description**
Describe your changes

**Testing steps**


ruanslv:
Results were not good (worse PPL with longer depth, lower WPS), so dropping this

PR #346, opened 2022-09-21
Add srctgt truncation

ramakanth-pasunuru:
- Add left or right truncation option for streaming_src_tgt_dataset module.

### DIFF
--- a/metaseq/data/streaming_src_tgt_dataset.py
+++ b/metaseq/data/streaming_src_tgt_dataset.py
@@ -39,6 +39,7 @@ class StreamingSrcTgtDataset(torch.utils.data.IterableDataset):
         drop_last: Optional[bool] = False,
         padding_idx: Optional[int] = None,
         shuffle_buffer_size: int = 1,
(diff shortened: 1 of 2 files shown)


metaseq/data/streaming_src_tgt_dataset.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         self.drop_last = drop_last
         self.padding_idx = padding_idx
         self.shuffle_buffer_size = shuffle_buffer_size
+        self.left_truncation = left_truncation
```
todpole3, line 52: Nit: renaming this to something like "--target-first-truncation"? I think three months later I will forget what "left_truncation" means lol

ramakanth-pasunuru: Actually, "target-first-truncation" is also confusing. Since we deal with left-right LMs, its implied that target is on the end of the block and the current name: `left_truncation` seems appropriate to me for *this* dataset module.

sriniiyer (approved):
lgtm

PR #352, opened 2022-09-27
Making data loader skipping 5--9x faster

zdevito:
The fast-forwarding process added with DeferredTensor avoids having to tokenize documents that have already been read, but the code to simulate the shuffle buffer and sequence construction process can still take ~7 minutes to execute on full-size shards.

This patch does some optimization to the hot path of the skipping process to make it somewhere between 5--9x faster compared to the current state and around 500 -- 900x faster than the original OPT method, which should bring the time to do the fast forward down below a minute.

* Merges the shuffling of documents, creation of sequences of documents into one object (DocumentToSequenceDataset). This allows all the deferred execution logic to be put into one file rather than have to expose a DeferredTensor object whose construction is slow.
* Uses coarse-grained per-worker locking rather than atomics so that we do not need C extensions anymore.
* Performs blocked generation of random numbers from numpy because there is high call overhead for asking for a single number.
* The merge of the datasets and change to the random numbers are done in such a  way that the random behavior of the merged object matches that of the previous code, so it is safe to have a checkpoint created in an older version of the code loaded by this code.
* I also manually check a requeue after checkpoint to see that the docsperex and loss match before and after checkpoint load.


cpu_tests/test_streaming_iterators.py
```diff
@@ -40,11 +41,41 @@ def get_simple_dataset():
...
 
 
+class FakeTensorData(torch.utils.data.Dataset):
+    def __init__(self):
+        self.rng = random.Random(0)
+        self.trng = torch.Generator()
+        self.trng.manual_seed(0)
+        self.items = [
+            torch.randint(
+                256, size=(self.rng.randrange(512, 513),), generator=self.trng
```
suchenzang, line 61: Nit: is `randrange(512, 513)` still needed here? Or meant to be kept at `randrange(512, 8192)` as before?

metaseq/data/document_to_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+        def documents():
+            for idx in indices:
+                ln = self.len_cache[idx]
+                if ln == 0:
+                    r = self.dataset[idx]
+                    ln = r.shape[0]
+                    self.len_cache[idx] = ln
+                    yield (ln, Cell(r))
+                else:
+                    create = (lambda i: lambda: self.dataset[i])(idx)
```
suchenzang, line 161: I'm assuming this is the crux of [previous comment to combine into single dataset object](https://github.com/facebookresearch/metaseq/pull/337/files#r976832368), but having a hard time following the indexing and nested lambdas. 😅   

Previously, `DeferredDataset` returned `DeferredTensor(ln, lambda: self.dataset[idx])` for `__getitem__`, but now.... we're indexing into different datasets?

cpu_tests/test_streaming_iterators.py
```diff
@@ -40,11 +41,41 @@ def get_simple_dataset():
             torch.LongTensor([8, 9]),
         ]
     )
-    dataset = DeferredDataset(dataset)
-    dataset = SkipDeferredDataset(dataset, 0)
+    dataset = DocumentToSequenceDataset(
+        dataset,
+        block_size=None,
+        permute_documents=False,
+        break_mode="passthrough",
```
stephenroller, line 48: This is a new breakmode. What is it? 😅

cpu_tests/test_streaming_iterators.py
```diff
@@ -227,17 +231,13 @@ def run_test(drop_last, break_mode):
...
-        defer_dataset.len_cache = len_cache
-        skip_dataset.to_skip = consumed
-        d = dataset
-        while hasattr(d, "dataset"):
-            if hasattr(d, "worker_offset"):
-                d.worker_offset = 7
-            d = d.dataset
+        token_dataset.len_cache = len_cache
+        token_dataset.to_skip = consumed
+        token_dataset.worker_offset = 7
```
stephenroller, line 240: Scary magic constant?!

cpu_tests/test_streaming_iterators.py
```diff
@@ -248,29 +248,65 @@ def run_test(drop_last, break_mode):
...
+            document_to_sequence_dataset = DocumentToSequenceDataset(
+                dataset,
+                # We generate blocks with one extra token, so that we have a target
+                # for the final input token. This results in slight data loss.
+                block_size=MAX_SEQ_LEN + 1,
+                break_mode=break_mode,
+                # we drop the remainder block during training
+                drop_last=drop_last,
+                padding_idx=1,
+                # 1284 is a randomly-generated offset to decouple the seed used here
```
stephenroller, line 283: Stale comment?

metaseq/data/document_to_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+import numpy as np
+import torch
+
+from metaseq.data import data_utils
+import time
+from metaseq.data.atomic_array import AtomicArray
+
+
+# gotta be a better way to get this...
+def f():
```
stephenroller, line 18: Wut

stephenroller (commented):
Hm is this going to break the fine tuning dataset? I need to look into that.

Going to give this a more thorough review later.

metaseq/data/document_to_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+        shuffle_buffer_size: int = 1,
+        seed: Optional[int] = None,
+        len_cache=None,
+        to_skip=0,
+        permute_documents=True,
+    ):
+        super().__init__()
+        self.dataset = dataset
+        assert len(dataset) > 0
+        self.worker_offset = 0
```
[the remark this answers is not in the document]
zdevito: Yeah, I will add a better comment here. What is happening:
* PyTorches data loader object pulls data round-robin from each worker, but it always starts with the _first_ worker when created, without any way to configure it.
* When we restart, we might have to restart by pulling data from a non-first worker (if n is not divisible by 8). Since we can't change how the PyTorch data loader object pulls data from the workers, we change the workers instead so that the first worker is the one we want to pull data from next.

zdevito:
> To confirm: do we expect ordering of data loaded to change after #337 and this PR? cc @stephenroller wondering if it's worth training another baseline out to 10k steps or so to confirm (mainly worried about accidentally shuffling / truncating within a dataset or breaking a document up in any weird way.. though if the "requeue from checkpoint" check above works, we should be ok?).

The expectation is the regardless of which patch, the data loading order should be the same. I agree that we should do some more rigorous testing to ensure that the changes didn't accidentally change any order. So far I have run it out 2500 iterations, reloaded the job and observed that loss/documents per eg stay exactly the same before and after. However, we probably also should check that before and after the patch the model trains the same way.

metaseq/data/document_to_sequence.py
zdevito: This is what is left of DeferredTensor after optimizing the Python by removing custom objects. 

`create` is a zero-argument function that actually performs the load of the tokens from a file. Python variable scoping rules require another function to be introduced to capture the _current_ value of `idx` int `i`, because`idx` will get updated to a new value along the next iteration of the loop.

Since Python has no compiler, optimized code ends up looking harder to understand. In particular, these lambdas/tuples represent the same thing that DeferredTensors did before but without the self-documenting nature of objects. I can add some comments to clarify.

cpu_tests/test_streaming_iterators.py
zdevito: A few of tests for saving/restoring state of iterators do not actually form sequences and just test loading single documents. Since I collapsed the skipping behavior into the SequencesToDocument object, I added "passthrough" to skip the construction of sequences and just return a whole document so those tests still do the same thing.

cpu_tests/test_streaming_iterators.py
zdevito: I cleaned this up

metaseq/data/document_to_sequence.py
zdevito: I was using this to get to Pythons 'cell' type which it uses for closed over variables, but I think this is extra confusing. I replaced the Cell's with single-element lists (at first  list[0] holds the `create` function to load token data, once loaded, list[0] holds the tokenized tensor).

metaseq/data/document_to_sequence.py
zdevito: I ended up reorganizing and adding some type annotations to remove the lambdas. Should be clearer now.

metaseq/data/document_to_sequence.py
```diff
@@ -0,0 +1,377 @@
...
+                yield get_next_item_and_replace_in_buffer(None)
+
+        for i, elem in enumerate(sequences()):
+            if i >= to_skip:
+                elem["ids"] = torch.LongTensor(elem["ids"])
+                subsequences = []
+                for c, start, ln in elem["block"]:
+                    # A padding tensor (<padding_value>, 0, length)
+                    if c[0] == "padding":
+                        subsequences.append(
```
suchenzang, line 251: Just to see if I follow from looking at type signatures: `elem` here is a `Sequence`, which means `elem["block"]` is a `List[SequenceFragment]`; the for loop here is then iterating through `c, start, ln` of `SequenceFragment`s, which means `c` is a `List[Document]`, which means `c[0]` is a `Document`?  

There's some magic with how `"padding"` state is being passed through possible block iterators (i.e. the `yield_*` methods?) so just wanting to confirm logic around this bit since `Document` annotation above now seems to be Union of `Tuple[...], str`..?

suchenzang (approved):
Just one outstanding question to confirm understanding - excited to test this out!!

metaseq/data/document_to_sequence.py
zdevito: I updated the documentation in the code to hopefully document this better, and document the rest of the file better.  Documents are a Union of int (index into the self.dataset of unloaded document, the string "padding" representing a document full of self.padding_idx, and torch.Tensor (already tokenized data). The Tuple thing was a bug in the annotations: the document iterator returns a tuple of `(size, [document])`. "padding" is the way the block iterators communicate that part of the sequence should be filled will padding. It is easiest to think of it conceptually as a tensor full of padding_idx of infinite size.

Please let me know what remains as the trickiest stuff to understand. I want to make sure this code stay maintainable while also being fast enough to do fast-forwarding without much noticeable cost.

metaseq/data/atomic_array.cpp
```diff
@@ -0,0 +1,235 @@
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+#include <iostream>
+
+static PyObject* kwnames;
+static PyObject* format;
+struct Owned {
+    Owned(PyObject* o)
+    : obj(o) {}
```
suchenzang, line 9: Will take this opportunity to learn about this syntax 😅 : what does `obj(o)` do here as part of the constructor member initialization?  And why is the destructor only decrementing the reference count for `obj`?

metaseq/data/iterators.py
```diff
@@ -332,11 +328,11 @@ def load_state_dict(self, state_dict):
                     distributed_utils.broadcast(len_tensor, 0, global_group)
                 else:
                     len_tensor = torch.empty(
-                        4 * len(dataset), dtype=torch.int8, device="cuda"
+                        4 * len(dataset.dataset), dtype=torch.int8, device="cuda"
```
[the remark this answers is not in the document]
zdevito: it's sizeof(int), I clarified.

metaseq/data/atomic_array.cpp
zdevito: Essentially it does `obj = o`. C++ has to provide an initial value to `obj` before the body of the function starts so that objects in the class otherwise all objects would need zero-argument default constructors.

zdevito:
I figured out a way to get rid of the C extension without losing performance. Patch should be simpler to follow now.

suchenzang:
> I figured out a way to get rid of the C extension without losing performance. Patch should be simpler to follow now.

Amazing - thanks for simplifying for us poor Pythonistas! 🙏

PR #356, opened 2022-09-29
Liliyu/bf16 init fix

lilisierrayu:
**Patch Description**
Adding fix to correctly initialize the non-MP model. And correctly infer inference datatype for model trained with mixed precision. 

**Testing steps**
Launch a consolidated model (mp=1) for a model trained with bf16.


stephenroller (approved):
Looks perfect, thanks for fixing for everyone!

PR #357, opened 2022-09-29
remove deprecated inferencre_step implementation

lilisierrayu:
**Patch Description**
"prefix_tokens" not used by sequence_generator or sequence_score, inference_step() implementation in lanague_modeling and language_modeling_inference_for_models_trained_with_streaming are the same as the one in base_task. 

**Testing steps**


stephenroller (approved):
ty!

PR #359, opened 2022-09-29
Changes flag

sriniiyer:
We need Changes=true during training for doc attention to work. We need Changes=False during eval. This solves this using an environment variable called EVAL.


stephenroller (changes requested):
Why don't you set it in inference mode at start? That's what the "make_generation_fast_" thing does.

stephenroller:
Or alternatively, why wouldn't we use the break mode arg to choose this logic?

todpole3 (approved):
lgtm

sriniiyer:
> Why don't you set it in inference mode at start? That's what the "make_generation_fast_" thing does.

Will look into this, using this for now to unblock others

PR #363, opened 2022-09-29
Improve regex for model loading

sriniiyer:
Fix the regex for model loading. Tested with 
```
CUDA_VISIBLE_DEVICES=0,1 FSD=~/data/instruct-opt/prompt_data/allbenchmarks_io_streaming_after_dedup_v5_sorted  python -m metaseq_internal.eval.gpt3_eval --tasks opti
 --opti-eval-set test_text_completion__ni_v2__task215_rocstories_incorrect_answer_generation__prompt0  --model-name opt_instr_test  --nb-few-shot-samples-values 0 --distributed-world-size 2 --fsdp --prediction
s-dump-dir ~/del_clf  --n-eval-samples 10
```

Output:
```
Loaded model
model_loading_time=1177.1 seconds
model_loading_time_cuda=1177.1 seconds
Using max-tokens=2048
task=opti
eval_set=test_text_completion__ni_v2__task215_rocstories_incorrect_answer_generation__prompt0
eval language=en
train_set=None
train_lang=None
template=optinstructbenchmark
calibration_options=[]
scoring=sum
nb_few_shot_samples=0
```

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -543,7 +543,7 @@ def load_model_ensemble_and_task(
         assert num_shards > 0
         for shard_idx in range(num_shards):
             if num_shards == 1:
-                filename = filename.replace(".pt", suffix + ".pt")
+                filename = re.sub('.pt$', suffix + ".pt", filename)
             else:
                 filename = orig_filename[:-3] + f"_part{shard_idx}.pt"
             if state is None:


metaseq/checkpoint_utils.py
```diff
@@ -543,7 +543,7 @@ def load_model_ensemble_and_task(
         assert num_shards > 0
         for shard_idx in range(num_shards):
             if num_shards == 1:
-                filename = filename.replace(".pt", suffix + ".pt")
+                filename = re.sub('.pt$', suffix + ".pt", filename)
```
tbmihailov, line 546: @sriniiyer  did you test if this still works for the old models?

stephenroller: Should we upstream this? How did we end up here?

sriniiyer: @tbmihailov  I tried this with 1.3B_gptz_azure and it works

@stephenroller We added a new flag to the checkpoint filename called pt_XX, so its mainly a problem for us. Upstream doesnt have this problem.

tbmihailov (approved):
LGTM!

PR #366, opened 2022-10-01
[Cleanup] Split out transformer components: 10.1/N

suchenzang:
Cleanup part... 10? Changes here include:
* removed `set_beam_size` and `reorder_incremental_state_scripting` methods in `incremental_decoder.py`
* moved `TransformerEncoder` and `Embedding` (moved to modules/) to separate files, renamed `transformer.py` -> `transformer_decoder.py`
* deleted unused `model_utils.py` file
* moved ffn and transformer encoder layer to separate files

Used a 125m baseline as a test run to confirm parity.


metaseq/modules/feedforward_network.py
```diff
@@ -0,0 +1,46 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from torch import nn as nn
+from torch.nn import functional as F
+
+from metaseq.modules import Linear, gelu
+from metaseq.modules.fused_bias_gelu import has_fused_bias_gelu, fused_bias_gelu
+
+
+def FeedForwardNetwork(x, fc1, activation_fn, fc2, dropout_module):
```
ruanslv, line 13: Nit: add a comment here with a high-level description, similar to the encode/decoder classes?

metaseq/modules/embedding.py
```diff
@@ -0,0 +1,29 @@
...
+from torch import nn as nn
+
+
+def Embedding(
+    num_embeddings,
+    embedding_dim,
+    padding_idx,
+    initialize_params_on_gpu=False,
+    dtype: Optional[torch.dtype] = None,
+):
```
ruanslv, line 18: Nit: add a comment here with a high-level description, similar to the encode/decoder classes?

ruanslv (approved):

PR #369, opened 2022-10-02
[Cleanup] Default to pre-norm throughput: 10.3/N

suchenzang:
[ Note: will rebase onto main after https://github.com/facebookresearch/metaseq/pull/367 and https://github.com/facebookresearch/metaseq/pull/366 are merged in ]

This PR removes *normalize_before flags and defaults to pre-norm logic throughout. 

Part of https://github.com/facebookresearch/metaseq/issues/368 efforts.

PR #370, opened 2022-10-02
[Cleanup] Remove encoder-related cruft in decoder code-paths: 10.4/N

suchenzang:
[ Note: this will be rebased on main after https://github.com/facebookresearch/metaseq/pull/369 is merged in (+ upstream diffs) ]

We currently default to `no_encoder_attn=True` in our decoder-only paths. Remove this cruft to cleanly separate between decoder and encoders.


suchenzang:
Ported into https://github.com/facebookresearch/metaseq/pull/381

PR #372, opened 2022-10-02
[Cleanup] Remove encoder_out/encoder_padding_mask from decoder stuff: 10.5/N

suchenzang:
[ Note this will be rebased onto main after https://github.com/facebookresearch/metaseq/pull/370 is merged ]

As title suggests, removing `encoder_out` and `encoder_padding_mask` args from decoder methods / classes.


suchenzang:
Ported into https://github.com/facebookresearch/metaseq/pull/381

PR #377, opened 2022-10-03
Meter logging fix

ramakanth-pasunuru:
This PR removes state dict of  HistoryMeter so that we do not write log of history information in this meter into checkpoints.

### DIFF
--- a/metaseq/logging/meters.py
+++ b/metaseq/logging/meters.py
@@ -77,22 +77,12 @@ class HistoryMeter(Meter):
         assert type(val) is list
         self.history.extend(val)
 
-    def state_dict(self):
-        return {
-            "history": self.history,
-            "round": self.round,
-        }
-
-    def load_state_dict(self, state_dict):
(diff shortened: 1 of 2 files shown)


metaseq/logging/meters.py
```diff
@@ -77,22 +77,12 @@ def update(self, val):
...
-        self.round = state_dict.get("round", None)
-
     @property
     def avg(self):
         if type(self.history[0]) is dict:
-            return sum([h["value"] for h in self.history])
+            return sum([h["value"] for h in self.history]) / len(self.history)
         else:
-            return sum(self.history)
+            return sum(self.history) / len(self.history)
```
sriniiyer, line 85: Is this related to removing the logging or is this a bug fix?

ramakanth-pasunuru: It's not related to logging. We never use it, but changed it anyway as per the module definition of returning some average.

sriniiyer (approved):
LGTM

PR #387, opened 2022-10-06
Add flash attention

dianaml0:
**Patch Description**
Describe your changes

**Testing steps**

### DIFF
--- a/metaseq/model_parallel/modules/transformer_layer.py
+++ b/metaseq/model_parallel/modules/transformer_layer.py
@@ -150,7 +150,7 @@ class ModelParallelTransformerDecoderLayer(TransformerDecoderLayer):
             num_layers=args.decoder_layers,
             dtype=utils.get_model_init_dtype(args),
             bias=not getattr(args, "disable_bias", False),
-            attn_variant=args.attn_variant,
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -66,15 +66,22 @@ install_fairscale: &install_fairscale
...
 
 install_dep_pt19: &install_dep_pt19
   - run:
       name: Install Pytorch Dependencies
       command: |
+        source $BASH_ENV
         source activate metaseq
         pip install --upgrade setuptools
-        pip install torch==1.9.1+cu111 torchvision==0.10.1+cu111 torchaudio==0.9.1 -f https://download.pytorch.org/whl/torch_stable.html
+        conda install pytorch torchvision torchaudio cudatoolkit=11.3 -c pytorch -q
```
suchenzang, line 83: Hmm I think this might be breaking the circleci GPU test

stephenroller (commented):
Let's make this an enum flag so we have future optionality on implementations. It will also let us eventually make the Megatron implementation an option (which people like to toggle at inference time)

stephenroller (approved):
Lgtm!

Issue #383, opened 2022-10-05
No longer able to load provided OPT checkpoint after recent changes


EIFY:
## 🐛 Bug

No longer able to load provided OPT checkpoint after recent changes

### To Reproduce

Edit `metaseq/service/constants.py` as before, in my case:
```python
MAX_SEQ_LEN = 2048
BATCH_SIZE = 2048  # silly high bc we dynamically batch by MAX_BATCH_TOKENS
MAX_BATCH_TOKENS = 3072
DEFAULT_PORT = 6010
MODEL_PARALLEL = 1
TOTAL_WORLD_SIZE = 1
MAX_BEAM = 16

try:
    # internal logic denoting where checkpoints are in meta infrastructure
    from metaseq_internal.constants import CHECKPOINT_FOLDER
except ImportError:
    CHECKPOINT_FOLDER = "/home/jason_chou/redspot_home/350m/"
(...)
```
where
```
$ pwd
/home/jason_chou/redspot_home
$ ls 350m/
dict.txt  gpt2-merges.txt  gpt2-vocab.json  reshard.pt
```
and then run `metaseq-api-local`, but it no longer works:
```
$ metaseq-api-local
2022-10-05 22:19:25 | INFO | metaseq.hub_utils | loading model(s) from /home/jason_chou/redspot_home/350m/reshard.pt
2022-10-05 22:19:26 | INFO | metaseq.checkpoint_utils | Done reading from disk
Traceback (most recent call last):
  File "/home/jason_chou/.conda/envs/user/bin/metaseq-api-local", line 8, in <module>
    sys.exit(cli_main())
  File "/home/default_user/metaseq/metaseq_cli/interactive_hosted.py", line 370, in cli_main
    distributed_utils.call_main(cfg, worker_main, namespace_args=args)
  File "/home/default_user/metaseq/metaseq/distributed/utils.py", line 279, in call_main
    return main(cfg, **kwargs)
  File "/home/default_user/metaseq/metaseq_cli/interactive_hosted.py", line 176, in worker_main
    models = generator.load_model()  # noqa: F841
  File "/home/default_user/metaseq/metaseq/hub_utils.py", line 565, in load_model
    models, _model_args, _task = _load_checkpoint()
  File "/home/default_user/metaseq/metaseq/hub_utils.py", line 548, in _load_checkpoint
    return checkpoint_utils.load_model_ensemble_and_task(
[...]

PR #391, opened 2022-10-06
Fix api model loading for mp=1

ruanslv:
**Patch Description**
Fixes part of https://github.com/facebookresearch/metaseq/issues/383. Model configs without MP don't have "inference" attribute, causing model load to fail with `omegaconf.errors.ConfigAttributeError: Key 'inference' is not in struct`.

"inference" field seems to be used only in MP for the code to track the current status. However, it is always set inside make_generation_fast_() (e.g. https://fburl.com/4gy1ufjv).

So we don't need this line, we can just rely on make_generation_fast_ to set inference to True for us when the model has MP enabled.

**Testing steps**
Loaded a model with and without MP (1.3B and 350M OPT).

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -535,7 +535,6 @@ class GeneratorInterface:
         task = tasks.setup_task(self.cfg.task)
 
         def _build_model(cfg, task):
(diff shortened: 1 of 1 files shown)


suchenzang (approved):
niiiice!

PR #395, opened 2022-10-07
Add GLU activations [v2]

ruanslv:
**Patch Description**
v2 for https://github.com/facebookresearch/metaseq/pull/343, with an attempt to decouple gate logic from decoder.

Also consolidated our GeLU implementation to a new version of gelu_accurate that explicitly defines the multiplying constants + relies on JIT for better performance.

**Testing steps**
Running ablations to compare perfomance against previous runs and making sure PPL matches


metaseq/modules/activation_functions.py
```diff
@@ -0,0 +1,78 @@
...
+def relu_squared(x: torch.Tensor):
+    return F.relu(x).pow(2)
+
+
+@torch.jit.script
+def gelu(x):
+    return x * 0.5 * (1.0 + torch.tanh(0.79788456 * x * (1 + 0.044715 * x * x)))
+
+
+def swiglu(x: torch.Tensor, gate: torch.Tensor):
```
stephenroller, line 17: JIT these please

metaseq/modules/feedforward_network.py
```diff
@@ -34,16 +34,18 @@ def FeedForwardNetwork(x, fc1, activation_fn, fc2, dropout_module):
...
         x = F.linear(x, fc1.weight, None)
         x = fused_bias_gelu(x, fc1.bias)
         x = F.linear(x, fc2.weight, fc2.bias)
     else:
-        x = fc1(x)
-        x = activation_fn(x)
+        fc1_out = fc1(x)
+        x = activation_fn(x, fc1_out, False)
+        del fc1_out
         x = fc2(x)
```
stephenroller, line 49: Instead of doing the manual deletion, let's just inline the call so it goes out of scope immediately

stephenroller: I guess that's awkward with the tuple. What is the second return value anyway?

ruanslv: It's the bias, see line 30. When skip_bias_add = True (which we set only when we want to use fused gelu), then the bias is not added to the weights and instead returned separately. In this case here, given we are adding bias the second return value is just None. Maybe there's a smart way to skip this

metaseq/modules/feedforward_network.py
```diff
@@ -34,16 +34,14 @@ def FeedForwardNetwork(x, fc1, activation_fn, fc2, dropout_module):
         x = x + bias_fc2
     elif model_parallel:
         # here, we do the bias computation inside fc1 and fc2 AND gather_output
-        x, _ = fc1(x)
-        x = activation_fn(x)
+        x = activation_fn(x, fc1(x)[0], True)
```
suchenzang, line 37: Nit: add `model_parallel=True` here - took me a good minute to figure out what this boolean meant

metaseq/modules/activation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+class ActivationFn(nn.Module):
+    def __init__(self, name, fc1_builder, embed_dim, ffn_dim, **fc1_kwargs):
+        super().__init__()
+        self.fn = self.__get_fn(name)
+        self.gate = None
+        if self.fn in self.__get_gated_fns():
+            self.gate = fc1_builder(embed_dim, ffn_dim, **fc1_kwargs)
+
+    def forward(self, fc1_in, fc1_out, model_parallel):
```
suchenzang, line 48: Nit: add a type annotation to `model_parallel` being a boolean flag

suchenzang (approved):
Just some nits

PR #398, opened 2022-10-07
Add support for source target style datasets to DocumentToSequence

zdevito:
Training runs using StreamingSrcTgtDataset were failing because they did not do the same token length caching as DocumentsToSequences.

StreamingSrcTgtDataset is really just another instances of StreamingTokenBlockDataset where the the blocks are split into a tuple (src, target). To avoid duplication this PR just adds support for this case directly to DocumentToSequences, and a test to verify this replicates the old behavior.

### DIFF
--- a/cpu_tests/test_streaming_iterators.py
+++ b/cpu_tests/test_streaming_iterators.py
@@ -62,7 +62,6 @@ class FakeTensorData(torch.utils.data.Dataset):
         self.queried = 0
         self.realized = [False for _ in self.items]
 
-
     def _gen_one(self, source_target):
         n = self.rng.randrange(512, 2048)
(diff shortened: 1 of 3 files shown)


zdevito:
@jxmsML

cpu_tests/test_streaming_iterators.py
```diff
@@ -53,19 +54,24 @@ def get_simple_dataset():
 
 
 class FakeTensorData(torch.utils.data.Dataset):
-    def __init__(self):
+    def __init__(self, source_target):
```
suchenzang, line 57: Nit: default value? Test fails on line 168

suchenzang (approved):

PR #406, opened 2022-10-13
Bugfix - Generation API fails if the generated sequence is truncated.

punitkoura:
**Patch Description**
1. Ensure that we pad the mask properly when calculating the token distribution.
2. Add a function to decode tokens in utils.py - This will be used internally.

**Testing steps**
Ran an API locally to verify generation works.

```
curl -k <URL> -H "Authorization: Bearer Punit" -H "Content-Type: application/json" \
> -d '{
> "prompt": [2, 44019, 28049, 13770, 6439, 22824, 2306, 14380, 6804, 5457, 21523, 50118, 50118, 713, 16, 10, 569, 9, 10, 11954, 6439, 3704, 81],
> "temperature": 0.0,
> "max_tokens": 300, "min_tokens": 0,
> "top_p": 1.0, "n": 1, "best_of": 1,
> "echo": true, "logprobs": 1, "seed": 1
> }'
```


davides:
Is there a unit test that would have caught this?

metaseq/service/utils.py
```diff
@@ -32,6 +33,14 @@ def encode_fn(generator, x):
     return generator.bpe.bpe.encode(normalize_newlines(x)).ids
 
 
+def decode_fn(generator, x: List[int]) -> str:
```
moyapchen, line 36: dumb q - where is this being used?

punitkoura: It will be used for evals internally... I could have added it there, but adding it here to keep related functions in the same place...

stephenroller: maybe it would be better to just move this to `generator` itself? Same with encode_fn

stephenroller (commented):
can we add a test?

metaseq/hub_utils.py
```diff
@@ -814,5 +814,9 @@ def _filter_special(
 
         # cut off at stop and drop pads
         if distributions is not None:
+            # If we broke early in the loop above, ensure that we
+            # fill mask with False upto distributions.shape[0]
+            mask.extend([False] * (distributions.shape[0] - len(mask)))
```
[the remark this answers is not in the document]
punitkoura: Added an assert to ensure this :)

metaseq/service/utils.py
punitkoura: Moved these functions to the generator

punitkoura:
Addressed all comments, added unit test to ensure this doesn't happen.

stephenroller (approved):

PR #411, opened 2022-10-14
Run evals during training

ruanslv:
**Patch Description**
Added a few config options so we can kick off an arbitrary evals command during training. This allows us to keep logic in OSS land without adding dependencies to our metaseq-internal evals code.

Instead of having the two command requirements, we could also add logic for them in metaseq directly, but they are cloud env specific, so it would mean more of our azure internal logic getting added to OSS land.

Instead, for now internally we can keep the logic in the eval scripts. 

**Testing steps**
Using our baseline sweeps setting the new params. Example:
`--eval_command = f"""
        python metaseq_internal/scripts/eval/schedule_jobs_few_shot_opt_evaluation.py
        -o {output_path} -t opt_eval_tasks --model-name {args.prefix}
        --model-template bf16_sharded_config
        --slurm-partition zetta
    """
 `


metaseq_cli/train.py
```diff
@@ -486,6 +494,36 @@ def post_checkpoint_callback(cfg, filename):
...
+                checkpoint_suffix,
+            )
+
+
+def _run_evaluations(eval_cmd, cloud_upload_path, local_file, checkpoint_suffix):
+    # WARNING: evals command needs to properly wait for
+    # all ranks to have finished uploading. Here we just
+    # wait for rank 0.
+    if distributed_utils.get_global_rank() != 0:
+        return
```
stephenroller, line 511: I think we need to figure this out, not just pray on rank0 to be done. We'll need this for NFS checkpointing too.

See here https://fb.workplace.com/groups/319878845696681/posts/2360341237454404

metaseq_cli/train.py
```diff
@@ -486,6 +494,36 @@ def post_checkpoint_callback(cfg, filename):
...
+    assert eval_cmd is not None, "--eval-command needs to be set."
+    cmd = eval_cmd.split()
+    checkpoint_name = local_file.split("/")[-1].replace(checkpoint_suffix, "")
```
stephenroller, line 514: are we sure we're not going to be triggering this on both checkpoint_last and checkpoint_X?

metaseq/dataclass/configs.py
```diff
@@ -498,6 +498,24 @@ class CheckpointConfig(MetaseqDataclass):
...
+            )
+        },
+    )
+    evaluate_interval_updates: int = field(
+        default=0, metadata={"help": "run eval_command every N updates"}
+    )
+    evaluate_last_checkpoint: bool = field(
+        default=False,
+        metadata={"help": "run the eval_command at the end of the training run"},
+    )
```
stephenroller, line 518: should we just tie this to validation?

metaseq/dataclass/configs.py
```diff
@@ -498,6 +498,24 @@ class CheckpointConfig(MetaseqDataclass):
         default=True,
         metadata={"help": "store a last checkpoint at the end of the training run."},
     )
+    eval_command: Optional[str] = field(
```
stephenroller, line 501: maybe post_validation_command?

metaseq/dataclass/configs.py
ruanslv: Sure, but I believe we are reaching 20+ tasks now. $$$$

metaseq_cli/train.py
ruanslv: Lol exactly the question I had, I'll give it a try (gloo + monitored_barrier)

metaseq_cli/train.py
ruanslv: checkpoint_last is only written at the end of the training run. checkpoint_X is written at the update intervals. The logic here is to figure out which name we should use for evals. 

I didn't fully get your concern, does this address it?

https://github.com/facebookresearch/metaseq/blob/43815b8d4b08933f0c2af1cd275ec9b0108e89ba/metaseq/checkpoint_utils.py#L82-L100

metaseq_cli/train.py
ruanslv: Added gloo pg + monitored_barrier

metaseq_cli/train.py
```diff
@@ -486,6 +500,43 @@ def post_checkpoint_callback(cfg, filename):
...
+    # The evals command must accept these two args and properly merge the cloud path
+    # with the checkpoint name.
+    cmd.extend(
+        [
+            "--model-cloud-path",
```
punitkoura, line 527: @ruanslv could you provide an example of what the cloud_upload_path and  checkpoint_name fields look like?

For context, on evals side, right now we're reusing the logic to specify a local model path like
```
FSD=/data/gpt-z/few_shot_data/  python metaseq_internal/eval/gpt3_eval.py \
--model-name 125m \
--tasks copa --nshot 0 --fsdp \
--results-dir ./local_results/testing \
--distributed-world-size 2 \
--model-path az://lrsstoragewest3/data/opt/gptzmodels/gptz/125M/reshard_no_os/reshard.pt \
--model-template gptz_sharded_config
```

ruanslv: Added a logger line for us to know which command is being run. Example:

`2022-10-26 20:58:11 | INFO | metaseq_cli.train | Kicking off evaluation cmd: ['python', 'metaseq_internal/scripts/eval/schedule_jobs_few_shot_opt_evaluation.py', '-o', './zucchini_evals/hookv', '-t', 'opt_eval_tasks', '--model-name', 'hookv', '--model-template', 'bf16_sharded_config', '--slurm-partition', 'zetta', '--model-path', '/data/gpt-z/models/zucchini_ablations/swiglu_fat/checkpoint_last.pt', '--model-cloud-path', 'https://lrsstoragewest3.blob.core.windows.net/ruanrms/checkpoints/hookv/?sv=2020-08-04&ss=bfqt&srt=sco&sp=rwdlacupitfx&se=2023-10-06T21:33:38Z&st=2022-05-09T13:33:38Z&spr=https&sig=lrvUOVNoS4p62JY3EV4XnNuJwq517ChOdFwitywQgeA%3D', '--model-cloud-filename', 'checkpoint_30.pt']`

ruanslv: Compared to your command, basically everything is set in the training script, apart from `--model-path` which we need to infer based on `--model-cloud-path` and `--model-cloud-filename`. 

It's not a simple concatenation given we still use complete Azure paths for training, which have the auth stuff at the end. Otherwise we could construct the full model-path here too

metaseq/dataclass/configs.py
stephenroller: Okay Sgtm but Option creep is historically one of the most frustrating things, so I look to always say “do we need that flag?”

metaseq_cli/train.py
stephenroller: Discussed offline about how it might be better to turn this into just a python function we import and call, and we start version controlling our eval stuff inside metaseq-internal.

metaseq_cli/train.py
```diff
@@ -486,6 +500,46 @@ def post_checkpoint_callback(cfg, filename):
...
+            cloud_upload_path,
+            "--model-cloud-filename",
+            checkpoint_name,
+        ]
+    )
+    logger.info(f"Kicking off evaluation cmd: {cmd}")
+    res = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if res.returncode != 0:
+        print("Error: {}, evaluation failed".format(res.returncode))
```
stephenroller, line 538: Logger.error instead?

metaseq_cli/train.py
```diff
@@ -486,6 +500,46 @@ def post_checkpoint_callback(cfg, filename):
...
+        print("Evaluation stdout = {}".format(res.stdout))
+        sys.exit(1)
```
stephenroller, line 540: Let’s not unexpectedly quit. Let’s raise an exception?

metaseq_cli/train.py
```diff
@@ -486,6 +501,34 @@ def post_checkpoint_callback(cfg, filename):
...
+):
+    # Make sure all ranks have finished uploading checkpoints.
+    # If any rank doesn't hit the barrier within the timeout period, we throw an error and do
+    # not run evals. Error doesn't stop training run.
+    dist.monitored_barrier(group=gloo_pg, timeout=timedelta(minutes=5))
+    # Run evals on rank 0
+    if distributed_utils.get_global_rank() != 0:
+        return
+    assert eval_module is not None, "--eval-module needs to be set."
+    module = importlib.import_module(eval_module)
```
stephenroller, line 524:
```suggestion
    module = importlib.import_module(eval_module)
    if not hasattr(module, 'eval_fn'):
         raise RuntimeError(
             f"{eval_module} must have a function called eval_fn to utilize for evaluations. "
             "It expects the following signature:\n"
             "def eval_fn(cloud_upload_path: str, checkpoint_name: str)"
         )
```

stephenroller (approved):
this looks good. let's just offer a more helpful error message if this is set wrong

ruanslv:
Local test working well, merging:

```
2022-11-02 20:01:05 | INFO | metaseq_cli.train | Kicking off eval_fn from: <module 'metaseq_internal.projects.zucchini.sweep_baseline' from '/shared/home/ruanrms/metaseq/metaseq-internal/metaseq_internal/projects/zucchini/sweep_baseline.py'>
args: https://lrsstoragewest3.blob.core.windows.net/ruanrms/checkpoints/hookfi/?sv=2020-08-04&ss=bfqt&srt=sco&sp=rwdlacupitfx&se=2023-10-06T21:33:38Z&st=2022-05-09T13:33:38Z&spr=https&sig=lrvUOVNoS4p62JY3EV4XnNuJwq517ChOdFwitywQgeA%3D, checkpoint_5.pt
2022-11-02 20:01:05 | INFO | metaseq_cli.train | Successfully ran evaluation
```

Issue #351, opened 2022-09-26
E2E tests for training resumption


ruanslv:
## 🚀 Feature Request
Adding tests for restarting a training run after some interruption

### Motivation
Restoring training runs from checkpoints is a critical feature for large-scale models and has multiple steps, including downloading checkpoint from blob storage, loading the data to the correct token, storing checkpoints after some conditions are met, etc.

Tests would allow us to safe-guard this critical flow in case of potential broken changes.

### Pitch

A CI test that trains a small model, stores a checkpoint file to blob, downloads that checkpoint file and successfully continues the training run.

### Alternatives

Unit-tests that cover each step separately.

### Additional context

https://github.com/facebookresearch/metaseq/commit/e3ea5070a8c1bae77703aef7fc0f5537bd437963 -- this commit caused some checkpoint files to stop being stored.

Xirider:
I would like to work on this.

suchenzang:
@Xirider As a heads up: you'll probably have to mock out the blob storage bit (implicitly testing our checkpointing logic too), so this would be quite involved (related issues: https://github.com/facebookresearch/metaseq/issues/268 and https://github.com/facebookresearch/metaseq/issues/142)

Issue #268, opened 2022-07-28
Add tests around checkpoint loading


suchenzang:
## 🚀 Feature Request
Add tests around loading checkpoints!

### Motivation

Right now we occasionally break codepaths related to checkpoint loading, since it isn't frequently tested internally until someone has to... load a checkpoint.  We will need to start adding some tests around this, since this area will also go through some refactoring to address https://github.com/facebookresearch/metaseq/issues/78

punitkoura:
We had added this test some time back to test checkpoint loading for a 125m model, and ensuring that the logits generated by the model parallel version and singleton version are the same.

We can make this more thorough to catch other errors as well...

suchenzang:
Another related issue: checkpoint downloading (this is at the intersection of two brittle components: checkpointing + our logic around different compute envs). https://github.com/facebookresearch/metaseq/pull/272

PR #426, opened 2022-10-17
E2E test for training resumption, mocks download and upload of azure checkpoints

Xirider:
Added E2E test for training resumption and azure checkpoint storing logic. Issue: https://github.com/facebookresearch/metaseq/issues/351 and https://github.com/facebookresearch/metaseq/issues/268
1. Training runs for 20 steps -> check for creation of the correct checkpoints and the correct uploads to azure blob and train loss
2. Training resumes for 15 steps -> check for correct "download" of last checkpoints from azure blob, load these checkpoints and have correct final loss

Azure blob download and upload is mocked out.

Note that newly spawned subprocesses do not keep mocked objects, so I had to instead pass functions that create the mocks inside the subprocesses.


suchenzang (approved):

suchenzang:
Yay tests!! Thank youu!

PR #431, opened 2022-10-17
add a double restart data loader test

zdevito:
This tests the change to worker offset (also small fix to the previous patch where a % was missing)


suchenzang (approved):
Thank youuu!!

PR #440, opened 2022-10-21
Add support for constants override

dgrnbrg-meta:
**Patch Description**
`constants.py` needs to be changed for every model configuration. Rather than editing the file in place, this change allows you to _optionally_ export an environment variable `METASEQ_SERVICE_CONSTANTS_MODULE` which is on the python path. When you do this, it will be used for the `interactive_hosted.py` configuration instead.

This allows you to have multiple, local flavors of constants that you can easily refer to.

**Testing steps**
I made a local override and successfully loaded a model that uses different `LAUNCH_ARGS` than what are commited.

### DIFF
--- a/metaseq/cli/README.md
+++ b/metaseq/cli/README.md
@@ -1,0 +1,63 @@
+# Easily running custom metaseq models
+
+Say you're running one of the interactive inference pragroms in `metaseq.cli`.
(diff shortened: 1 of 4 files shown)


suchenzang:
Nit on adding a smol readme for how to add this override / demo this?

suchenzang (approved):

Ansl55555 (commented):
>

Issue #236, opened 2022-07-19
Make DATA_LOCATIONS for training easily configurable for external users


pragnakalpdev27:
#### What is your question?
We are trying to train OPT-125m model for our custom dataset using Google Colab pro with one GPU.
Which files need to store in DATA_LOCATIONS?

#### Code

! python metaseq/launcher/opt_baselines.py \
  -n 1 -g 2 \
  -p test_v0 \
  --model-size 125m \
  --azure \
  --data /content/test_data \
  --checkpoints-dir "/content/ck_point" \
  --no-tensorboard \

#### What have you tried?
We are facing the below error while running the baseline command. We have updated path in metaseq/launcher/opt_job_constants.py file in below line:

DATA_LOCATIONS = { 
     ComputeEnvs.AZURE: "/data/opt", 
 } 

But we are facing an error as shown in below:

Traceback (most recent call last):
  File "/usr/local/bin/opt-baselines", line 33, in <module>
    sys.exit(load_entry_point('metaseq==0.0.1', 'console_scripts', 'opt-baselines')())
  File "/usr/local/lib/python3.7/dist-packages/metaseq-0.0.1-py3.7-linux-x86_64.egg/metaseq/launcher/opt_baselines.py", line 337, in cli_main
    get_grid, postprocess_hyperparams, add_extra_options_func=add_extra_options_func
  File "/usr/local/lib/python3.7/dist-packages/metaseq-0.0.1-py3.7-linux-x86_64.egg/metaseq/launcher/sweep.py", line 378, in main
    backend_main(get_grid, postprocess_hyperparams, args)
  File "/usr/local/lib/python3.7/dist-packages/metaseq-0.0.1-py3.7-linux-x86_64.egg/metaseq/launcher/slurm.py", line 34, in main
    grid = get_grid(args)
  File "/usr/local/lib/python3.7/dist-packages/metaseq-0.0.1-py3.7-linux-x86_64.egg/metaseq/launcher/opt_baselines.py", line 88, in get_grid
    raise RuntimeError("Where are you running this?! Check DATA_LOCATIONS.")
RuntimeError: Where are you running this?! Check DATA_LOCATIONS.

#### What's your environment?

  - metaseq Version (e.g., 1.0 or master): Forked Repo
 - PyTorch Version (e.g., 1.0) :  1.11.0
 - OS (e.g., Linux): Linux 
 - How you installed metaseq (`pip`, source): source
 - Build command you used (if compiling from source):
[...]

PR #443, opened 2022-10-24
Allow user to specify external data paths

Xirider:
Quick fix to allow users outside of the specified compute envs to specify a data path and avoid the hard coded folder names until we rework the config system.
Solves https://github.com/facebookresearch/metaseq/issues/236

### DIFF
--- a/metaseq/launcher/opt_baselines.py
+++ b/metaseq/launcher/opt_baselines.py
@@ -80,23 +80,24 @@ def get_grid(args):
     # Infer data path if not given
     DATA_ROOT = ""
     valid_subsets = VALID_SUBSETS
-    if args.data is None and not args.benchmark:
-        cluster_env = get_env_from_args(args)
-        data_loc_by_env = DATA_LOCATIONS[cluster_env]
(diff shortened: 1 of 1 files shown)


metaseq/launcher/opt_baselines.py
```diff
@@ -90,7 +90,13 @@ def get_grid(args):
         if os.path.exists(args.data):
             DATA_ROOT = data_loc_by_env
         else:
-            raise RuntimeError("Where are you running this?! Check DATA_LOCATIONS.")
+            raise RuntimeError(
+                "Where are you running this?! Check DATA_LOCATIONS or pass --data "
+                "pointing to a directory with 'data' and 'tokenizers' folders."
+            )
+    elif not args.benchmark:
```
suchenzang, line 97: Nit: move this conditional to top-level to make this branching easier to read? Something like:

```
if not args.benchmark:
    if args.data is None:
    ...
    else:
    ...
```

suchenzang (approved):
Thanks for fixing this papercut!!

PR #449, opened 2022-10-24
Remove SequenceScorer

punitkoura:
**Patch Description**
Delete SequenceScorer from metaseq, we are not using it anymore. Use SequenceGenerator now.

**Testing steps**
Regression tests for hub_utils.py


metaseq/tasks/base_task.py
```diff
@@ -338,15 +338,6 @@ def build_criterion(self, cfg: DictConfig):
     def build_generator(
         self, models, args, seq_gen_cls=None, extra_gen_cls_kwargs=None
     ):
-
-        if getattr(args, "score_reference", False):
```
punitkoura, line 342: We no longer use this kwarg

suchenzang (approved):
🔥

Issue #437, opened 2022-10-19
Wrong mode for opening file when profiling when invoked with opt-baselines


yosoyjay:
## 🐛 Bug

When invoking use of the profiler (e.g. `opt-baselines ... --profile`) a file `memory_usage.txt` is opened in read mode and very likely does not exist.

### To Reproduce

`opt-baselines --model-size 8m --benchmark -t 1 -g 8 -n 1 -p profile-memory-file-bug --azure --local --profile`

#### Code sample

```
opt-baselines --model-size 8m --benchmark -t 1 -g 8 -n 1 -p profile-memory-file-bug --azure --local --profile
...
Traceback (most recent call last):
  File "/shared/home/hpcadmin/./slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq_cli/train.py", line 624, in <module>
    cli_main()
  File "/shared/home/hpcadmin/./slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq_cli/train.py", line 620, in cli_main
    distributed_utils.call_main(cfg, main)
  File "/shared/home/hpcadmin/slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq/distributed/utils.py", line 272, in call_main
    return _spawn_helper(main, cfg, kwargs)
  File "/shared/home/hpcadmin/slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq/distributed/utils.py", line 250, in _spawn_helper
    retval = distributed_main(-1, main, cfg, kwargs)
  File "/shared/home/hpcadmin/slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq/distributed/utils.py", line 212, in distributed_main
    retval = main(cfg, **kwargs)
  File "/shared/home/hpcadmin/./slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq_cli/train.py", line 174, in main
    valid_losses, should_stop = train(cfg, trainer, task, epoch_itr)
  File "/shared/home/hpcadmin/micromamba/envs/fairseq/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/shared/home/hpcadmin/./slurm_snapshot_code_oss/2022-10-19T23_09_03.572950/metaseq_cli/train.py", line 339, in train
    print(
io.UnsupportedOperation: not writable
```

### Expected behavior

[...]

PR #450, opened 2022-10-25
fixed issue 437 and removed retired profiler

igormolybogFB:
**Patch Description**
the flag --profile in metaseq train.py was not accessible through opt_baseline.py or sweep_baseline.py that we normally use. This is due to --profile flag in those was mapped to --new_profiler flag in train.py and not the --profile flag. 

Here is the description of how that is happening
(How profiler option gets into slurm job):
1. [sweep_baseline.py] –profile [is read](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/projects/zucchini/sweep_baseline.py#L75) from cli
2. –new-profile [is added](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/projects/zucchini/sweep_baseline.py#L195) to the grid (profile is not)
3. [slurm.py] [config](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/fb_sweep/slurm.py#L490) is produced from the grid
4. train_cmd [is extended](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/fb_sweep/slurm.py#L225) from config
5. srun_cmd [is extended](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/fb_sweep/slurm.py#L290) from train_cmd (and srun_cmd_str)
6. srun_cmd_str -> wrapped_cmd -> sbatch_cmd 
run_batch (sbatch_cmd) is [called](https://github.com/fairinternal/metaseq-internal/blob/main/metaseq_internal/fb_sweep/slurm.py#L578)

Moreover, --profile flag corresponds to the outdated version of profiler (torch.autograd.profiler) and not the new one (torch.profiler). As per @ngoyal2707 request, the outdated profiler gets cleaned out of our code (in metaseq only) and --profile in both metaseq-internal and metaseq are being matched.

Besides that, [issue 437](https://github.com/facebookresearch/metaseq/issues/437) is fixed by implementing the suggestion

**Testing steps**
run 

python -m metaseq_internal.projects.zucchini.sweep_baseline -g 2 -n 1 --azure --model-size 125m --data /data/gpt-z/zucchini/consolidated/v1.0 --tokenizer noregex --partition zetta --prefix profile_run --profile


suchenzang (approved):
🧹 🧹 

(just a nit on lint)

anupambhatnagar:
@igormolybogFB Can we get a screenshot of how the trace looks like with the new profiler? It would be helpful to see if we can see the python call stack when the `with_stack` is set to True. With an earlier version of kineto we encountered segfault when trying to collect the trace, so we should confirm that it is no longer the case.

igormolybogFB:
@anupambhatnagar 
Screen Shot 2022-10-26 at 3 35 02 PM

PR #452, opened 2022-10-25
Simplify checkpoint download logic

davides:
**Patch Description**
@moyapchen ran into a FileNotFoundError in checkpoint_utils while evaluating the 125M model with MP=2. I haven't been able to repro it consistently but I suspect that the issue is related to race condition in the force-download logic here: https://github.com/facebookresearch/metaseq/blob/main/metaseq/checkpoint_utils.py#L374-L389

This change simplifies it by just passing `force=True` to `get_local_path()`, which tells the handler to ignore caching rules and re-download the target file.

**Testing steps**
Multiple runs of the 125M eval command that first caught the issue. 

Also added a utility to stress test the file locking logic:
```
python -m tests.file_io.async_download_test
```


metaseq/checkpoint_utils.py
```diff
@@ -327,31 +326,77 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
...
+        Input: path=remote://path/checkpoint_last-model_part-0.pt, suffix=shard
+        Output: [/local/checkpoint_last-model_part-0.pt]
+
+    Ex. 3:
+        Remote: [path://checkpoint_last.pt]
+        Input: path=remote://path/checkpoint_last-shard0.pt, suffix=shard
+        Output: []
+    """
+    src_dir = os.path.dirname(path)
+    path_prefix = re.sub(rf"{suffix}[0-9]+\.pt$", f"{suffix}*", path)
```
punitkoura, line 352: Could you add a comment here for what this checks? Looks like this is handling the scenario when we don't have shards, just model part files?

punitkoura: Maybe an example similar to the examples in the docstring

metaseq/checkpoint_utils.py
```diff
@@ -327,31 +326,77 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
...
+
+    return local_paths
+
+
+def get_paths_to_load(path, suffix="rank-"):
+    checkpoint_files = _cache_checkpoint_files(path, suffix)
+
+    # Check if this looks like a sharded checkpoint
+    if not _is_checkpoint_sharded(checkpoint_files):
+        return [checkpoint_files[0]]
```
punitkoura, line 375: Should we assert here that checkpoint_files is a singleton list?

metaseq/checkpoint_utils.py
```diff
@@ -327,31 +326,77 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
...
+
+    # Check if we actually need to assign >1 shard
     checkpoint_files_count = len(checkpoint_files)
     world_size = distributed_utils.get_data_parallel_world_size()
-    fnames = []
     if world_size >= checkpoint_files_count:
-        return [local_path]
+        return [checkpoint_files[0]]
```
punitkoura, line 381: Same as above

punitkoura:
LGTM

tests/file_io/async_download_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def get_base_path():
+    base_path = os.environ.get(ENV_BASE_PATH, None)
+    assert base_path is not None, f"Required env var not specified: {ENV_BASE_PATH}"
+    # Set a unique base path for each test run
+    return os.path.join(base_path, str(uuid.uuid4()))
+
+
+# def get_local_test_file(parent_dir, name, length=4096):
```
moyapchen, line 39: Nit: remove commented code?

suchenzang:
I think a rebase on main should hopefully fix the failing test...

stephenroller (commented):
Cc @zdevito worth recognizing these changes

punitkoura (approved):
LGTM, thanks for fixing this!

PR #453, opened 2022-10-25
Fix stopping during generation

sriniiyer:
t is a tensor, and t in set() does not work. The correct way is to use t.item().


ramakanth-pasunuru (approved):
LGTM

PR #459, opened 2022-10-26
Add a script to reshard FSDP checkpoints

tangbinh:
### Summary
We add a new script to reshard raw FSDP checkpoints as part of our efforts to consolidate the checkpoint resharding logic. This script is a bit more general than some of the existing ones:
- Compared to [reshard_mp.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_mp.py), it allows us to optionally unflatten model weights and be compatible with the generator interface when `ddp-backend` is set to `pytorch_ddp`.
- Compared to the [`consolidate_shard_weights`](https://github.com/facebookresearch/fairscale/blob/ee647b976cf4c8fdd37bc9ae3fd6331d225ba2a0/fairscale/nn/data_parallel/fully_sharded_data_parallel.py#L2161) and [`build_unflat_state_dict`](https://github.com/facebookresearch/fairscale/blob/ee647b976cf4c8fdd37bc9ae3fd6331d225ba2a0/fairscale/nn/data_parallel/fsdp_optim_utils.py#L200) functions from FSDP (the former is used in [stitch_fsdp_ckpt.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/distributed/stitch_fsdp_ckpt.py#L152)), it supports both unsharding and resharding model weights and optimizer states.
- Compared to [checkpoint_utils.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/checkpoint_utils.py#L428), which is used in [`convert_to_singleton.py`](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/convert_to_singleton.py#L106-L124), it doesn't require instantiating FSDP instances and avoid the various requirements that come with it (DDP, vocab files, configs, etc). We also decouple the filename handling to make it a bit more flexible.

Note that this script doesn't include the logic for model parallel resharding. We should probably have a separate script for it, which can be used together with this one.

### Testing

- Run the script to merge the sharded checkpoints of the 2.7B parameters model into one shard for each model parallel part and load the resharded checkpoints with the interactive CLI:
```bash
for j in {0..3}; do
    python -m metaseq.scripts.reshard_fsdp \
    --input-glob-pattern "/data/gpt-z/models/gptz/2.7B/raw/checkpoint_last-model_part-$j-shard*.pt" \
    --output-shard-name "/shared/home/binhtang/checkpoints/opt-2.7b/reshard-model_part-$j.pt" \
    --num-output-shards 1 --skip-optimizer-state True --unflatten-weights True;
done
```
```bash
python -m metaseq.cli.interactive_cli
> what is the meaning of life?
To be happy.
```

- Run the script to reshard the 6.7B parameters model checkpoint for each model parallel part from 256 shards to 1 shard and from 1 shard back to 256 shards. The sharded checkpoints we get back are almost identical to the original ones except for some rank-specific data that are lost during the first conversion due to rank 0 copies (e.g `optimizer_history`, `extra_state`, `cfg.distributed_training.distributed_rank`). 
```bash
for j in {0..1}; do
    python -m metaseq.scripts.reshard_fsdp \
    --input-glob-pattern "/data/gpt-z/models/gptz/6.7B/raw/checkpoint_last-model_part-$j-shard*.pt" \
    --output-shard-name "/shared/home/binhtang/checkpoints/opt-6.7b/reshard-model_part-$j.pt" \
    --num-output-shards 1 --skip-optimizer-state False --unflatten-weights False;
done

for j in {0..1}; do
    python -m metaseq.scripts.reshard_fsdp \
    --input-glob-pattern "/shared/home/binhtang/checkpoints/opt-6.7b/reshard-model_part-$j.pt" \
    --output-shard-name "/shared/home/binhtang/checkpoints/opt-6.7b-reshard/checkpoint_last-model_part-$j-shard{i}.pt" \
    --num-output-shards 256 --skip-optimizer-state False --unflatten-weights False;
done
```

```python
import torch
for i in range(256):
    before = torch.load(f"/data/gpt-z/models/gptz/6.7B/raw/checkpoint_last-model_part-0-shard{i}.pt", map_location=torch.device("cpu"))
    after = torch.load(f"/shared/home/binhtang/checkpoints/opt-6.7b-reshard/checkpoint_last-model_part-0-shard{i}.pt", map_location=torch.device("cpu"))
    assert all(torch.allclose(before["model"][k], after["model"][k]) for k in before["model"].keys())
    assert(before["shard_metadata"] == after["shard_metadata"])
    assert(torch.allclose(x['exp_avg'], y['exp_avg']) for x, y in zip(before['last_optimizer_state']['state'], after['last_optimizer_state']['state']) for key in ('exp_avg', 'exp_avg_sq'))
```


suchenzang:
Before digging in... are we now able to delete at least one of the existing resharding/consolidate scripts 😅 ?

tangbinh:
> Before digging in... are we now able to delete at least one of the existing resharding/consolidate scripts 😅 ?

I'm not sure about the deprecation practice for our repo, but I think we can deprecate [reshard_mp.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_mp.py) and its related Bash scripts ([reshard_mp_launch.sh](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_mp_launch.sh) and [reshard_mp_launch_no_slurm.sh](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_mp_launch_no_slurm.sh)) after this PR. The introduced script also overlaps heavily with some functions in [checkpoint_utils.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/checkpoint_utils.py), and I wonder if it makes sense to combine them as well (perhaps after more thorough testing).

The other scripts ([reshard_model_parallel.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_model_parallel.py), [convert_to_singleton.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/convert_to_singleton.py), [consolidate_fsdp_shards.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/consolidate_fsdp_shards.py), [stitch_fsdp_ckpt.py](https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/distributed/stitch_fsdp_ckpt.py)) have the MP resharding logic and can be addressed later on (I'm working on another script for MP resharding only). What do you think?

suchenzang:
I would've just yolo deleted but having a semblance of deprecation practice here seems like the right thing to do - would just move all the ones we can deprecate now into some deprecated directory, then delete them in a few months / whenever we remember to clean things out again, lol.

suchenzang:
Since I'm depending mostly the test plan to stamp this... can you add a quick update to a README somewhere mentioning how / when to use this script (and a placeholder for the model parallel resharding later on)?

tangbinh:
> would just move all the ones we can deprecate now into some deprecated directory, then delete them in a few months
Good idea! Perhaps we can also add some logging statements in those scripts to let people know about the deprecation?

> can you add a quick update to a README somewhere mentioning how / when to use this script
I'll add more docstring in this script then. It doesn't seem to fit well into any of the existing READMEs.

metaseq/scripts/reshard_fsdp.py
```diff
@@ -0,0 +1,235 @@
...
+        resharded_state_dicts[shard_idx]["model"] = weight
+        resharded_state_dicts[shard_idx]["shard_metadata"] = metadata
+
+    # Unshard last optimizer state
+    if not skip_optimizer_state and "last_optimizer_state" in shard_state_dicts[0]:
+        reshared_optim_states = reshard_fsdp_optim_state(
+            [state["last_optimizer_state"] for state in shard_state_dicts],
+            num_output_shards,
+        )
+        # TODO: Support optimizer state unpadding
```
ruanslv, line 93: Will script work without this TODO? Which use-case is missing?

tangbinh: This part is similar to the existing [reshard_mp.py](https://github.com/facebookresearch/metaseq/blob/5ca6005552b7a049f071754393d5b86185e0db6a/metaseq/scripts/reshard_mp.py#L141) script where no unpadding is applied. The other script [checkpoint_utils.py](https://github.com/facebookresearch/metaseq/blob/5ca6005552b7a049f071754393d5b86185e0db6a/metaseq/checkpoint_utils.py#L660) does have some unpadding logic, but I'm not entirely sure if it's correct since it tries to find the padding info for the optimizer states from that of the model parameters and presumably always get 0 padding. I need to double check if we actually save the padding info for the optimizer states in the existing checkpoints.

metaseq/scripts/reshard_fsdp.py
```diff
@@ -0,0 +1,235 @@
...
+    Args:
+        :param input_glob_pattern: A glob pattern specifying the path names of the input shards.
+            (e.g. "checkpoints/2.7B/raw/checkpoint_last-model_part-0-shard*.pt").
+        :param output_shard_name: A string pattern specifying the path names of the output shards.
+            Shard indices can be included in the path names if the pattern includes `{i}`.
+            (e.g. "checkpoints/2.7B/reshard/reshard-model_part-0-shard{i}.pt").
+        :param num_output_shards: The number of output shards.
+        :param unflatten_weights: Specifies whether the weights in the input shards should be
+            unflattened. This option is only applicable when the number of output shards is 1.
+        :param skip_optimizer_state: Specifies whether to skip the optimizer states from the input shards.
```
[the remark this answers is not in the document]
tangbinh: This option can be useful as unless we want to resume training, the optimizer states aren't always needed (e.g. at inference time). They take up a lot of space and may result in issues under RAM constraints. For example, the 6.7B model is only 12.6 GB but with the optimizer states, it bloats to 52 GB. This option was also available in some of the existing scripts (e.g. see [reshard_mp.py](https://github.com/facebookresearch/metaseq/blob/5ca6005552b7a049f071754393d5b86185e0db6a/metaseq/scripts/reshard_mp.py)).

tangbinh:
> Could we also test this with output shards > 1? Even better if output shards > input shards, which is something we'll need for big zucchini run

Please see the updated test plan for the 256 shards &rarr; 1 shard &rarr; 256 shards conversion. In contrast to the previous scripts that [throw away sharded metadata](https://github.com/facebookresearch/metaseq/blob/5ca6005552b7a049f071754393d5b86185e0db6a/metaseq/scripts/reshard_mp.py#L116), we're able to chain these resharding operations by updating the metadata after each step.

metaseq/scripts/reshard_fsdp.py
ruanslv: So just to confirm my understanding, padding doesn't affect the final checkpoint files, it just increases storage size a bit? So we can live without this optimization for now?

suchenzang:
One last sanity check: could we check that the 256 shard -> 1 shard conversion yields a model that generates the same output given the same input?  And same for 1 shard -> 256 shard (need some... numerical equivalence checks to make sure we didn't accidentally jumble up the parameters 😅 ). In one of these cases, would also be good to check when we include optimizer state with a reshard operation and confirm that we can successfully resume training from a resharded checkpoint (and that training continues to look "sane").

We can separately note the above in a separate task to tackle and just merge this PR to wrap up this bit for now.  The above testing would be super important to confirm / write a test case around, but don't want to block on that.

metaseq/scripts/reshard_fsdp.py
tangbinh: Ah, I've confirmed with Naman about the padding logic for the optimizer states. This part is now completed and tested.

tangbinh:
Training appears to continue as expected after we load the resharded checkpoints. For example, here is the loss curve after we shard a checkpoint for the 125M model from 4 shards to 8 shards and reload the new checkpoint from step 2000:
Screen Shot 2022-11-11 at 9 25 43 AM

suchenzang (approved):
:shipit:

PR #464, opened 2022-10-28
Fix extra valid at beginning

sriniiyer:
validate_and_save() already checks for validate_at_beginning, so we don't need this extra call to validate.

See Line 422


todpole3 (approved):
Ty!

PR #466, opened 2022-10-28
Add full_megatron_init_scalar flag

suchenzang:
To scale fc2 in FFN and out_proj in MHA with.

### DIFF
--- a/metaseq/model_parallel/modules/transformer_layer.py
+++ b/metaseq/model_parallel/modules/transformer_layer.py
@@ -96,12 +96,14 @@ class ModelParallelTransformerDecoderLayer(TransformerDecoderLayer):
         )
 
     # TODO[susanz]: unify method signatures with non-model-parallel version.
+    # Note: only fc2 includes the full_megatron_init_scalar arg, given the scaled_init_method_normal call.
     def build_fc2(
         self,
         input_dim,
         output_dim,
         initialize_params_on_gpu,
         full_megatron_init,
+        full_megatron_init_scalar,
         megatron_init_sigma,
         num_layers,
         dtype,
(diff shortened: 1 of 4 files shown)


metaseq/model_parallel/modules/transformer_layer.py
```diff
@@ -110,7 +112,7 @@ def build_fc2(
         skip_bias_add = self.skip_bias_add
         if full_megatron_init:
             init_method_weights = megatron_utils.scaled_init_method_normal(
-                megatron_init_sigma, num_layers
+                megatron_init_sigma, num_layers * full_megatron_init_scalar
```
Xirider, line 115: Nit: It might be more straightforward to scale `megatron_init_sigma` instead?

metaseq/models/transformer_lm.py
```diff
@@ -132,6 +132,12 @@ class TransformerLanguageModelConfig(MetaseqDataclass):
         default=False,
         metadata={"help": "Exact same init as Megatron"},
     )
+    full_megatron_init_scalar: float = field(
```
Xirider, line 135: Nit: An arg name that includes second_layer makes might be more understandable: `full_megatron_init_scalar_second_layer`

Xirider (approved):

metaseq/model_parallel/modules/transformer_layer.py
suchenzang: Nah, want to keep that constant since that affects other layers too.

metaseq/models/transformer_lm.py
suchenzang: It includes MHA out proj so.. it's sadly kind of misnamed either way :(

metaseq/model_parallel/modules/transformer_layer.py
Xirider: The thought here was:
```
init_method_weights = megatron_utils.scaled_init_method_normal(
    megatron_init_sigma * full_megatron_init_scalar, num_layers )
```

suchenzang: Ah good callout - that does seem a lot more straightforward; I'll amend!

PR #473, opened 2022-11-01
Added dynamic configs (profiling)

igormolybogFB:
**Patch Description**
Made tombstoning work for RSC:

Added dynamic configs:

enabling:
 Just include  --dynamic-config-path and a specified json file path. The file must exist and will be picked up as the default dynamic config file.
 
using: 
1. when you want to change the value of a dynamic configuration just amend it in the dynamic config json file and save 
2. Allow some time for the changes to propagate (default timeout is 30 sec)

how it works: 
dynamic configuration is a dict with built-in timeout. If you are trying to access a value in the dict before the time is out, it will just give you the value with minimal overhead. If you are trying to access a value after the timeout, the entire dict will be reloaded from file and timer reset.

why it is useful:
If there is some computationally expensive logging/profiling that needs to be done only when weird behavior of the training procedure is observed, one should be able to trigger these operations on demand.

I have added the first dynamic config flag "force_profile" - it enables enabling profiling not only on step 5 but anywhere throughout training, even if cfg.common.profile = False

**Testing steps**

(using metaseq-internal) on Azure, run:
python -m metaseq_internal.projects.zucchini.sweep_baseline -g 8 -n 1 --azure --model-size 125m --data /data/gpt-z/zucchini/consolidated/v1.0 --tokenizer noregex --partition zetta --profile --dynamic-config --prefix dcfg_test

Results are in 
/shared/home/igormolybog/checkpoints/dcfg_test22/
(including traces from mnt/)

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -30,6 +30,7 @@ from metaseq import (
 )
 from metaseq.data import iterators, data_utils
 from metaseq.data.plasma_utils import PlasmaStore
+from metaseq.dataclass.configs import DynamicConfig
 from metaseq.dataclass.utils import convert_namespace_to_omegaconf
 from metaseq.distributed import fsdp_enable_wrap, fsdp_wrap, utils as distributed_utils
 from metaseq.file_io import PathManager
@@ -38,7 +39,7 @@ from metaseq.model_parallel.megatron_trainer import MegatronTrainer
 from metaseq.trainer import Trainer
 
 logging.basicConfig(
-    format="%(asctime)s | %(levelname)s | %(name)s | %(message)s",
+    format=f"slurm_procid {os.environ['SLURM_PROCID']} : %(asctime)s | %(levelname)s | %(name)s | %(message)s",
     datefmt="%Y-%m-%d %H:%M:%S",
     level=os.environ.get("LOGLEVEL", "INFO").upper(),
     stream=sys.stdout,
@@ -308,9 +309,18 @@ def train(
 
         return valid_losses, should_stop
 
+    dcfg = DynamicConfig(
+        json_file_path=cfg.common.dynamic_config_path,
+        timeout=cfg.common.dynamic_config_timeout,
+    )
+
     for i, samples in enumerate(progress):
-        if distributed_utils.get_global_rank() == 0 and cfg.common.profile and i == 5:
-            logger.info("STARTING PROFILER")
+        force_profile = dcfg["force_profile"]
+        do_profile = (distributed_utils.get_global_rank() == 0) and (
+            (cfg.common.profile and i == 5) or force_profile
+        )
+        if do_profile:
+            logger.info(f"STARTING PROFILER: step {i}")
             with profiler.profile(
                 profile_memory=True, with_stack=True, record_shapes=True
             ) as prof:
@@ -326,8 +336,9 @@ def train(
                     file=sourceFile,
                 )
             prof.export_chrome_trace(
-                os.path.join(cfg.checkpoint.save_dir, "profiler_trace.json")
+                os.path.join(cfg.checkpoint.save_dir, f"profiler_trace_step_{i}.json")
             )
+            logger.info(f"FINISHING PROFILER: step {i}")
         else:
             valid_losses, should_stop = train(i, samples)
         if should_stop:
--- a/metaseq/dataclass/configs.py
+++ b/metaseq/dataclass/configs.py
@@ -4,6 +4,9 @@
 # LICENSE file in the root directory of this source tree.
 
 import sys
+import time
+import json
+import os
 from dataclasses import _MISSING_TYPE, dataclass, field
 from typing import Any, List, Optional
 
@@ -193,6 +196,15 @@ class CommonConfig(MetaseqDataclass):
     )
     log_nvidia_smi: bool = field(
         default=False, metadata={"help": "log output from nvidia-smi during training"}
+    )
+    dynamic_config_path: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": "a path to place a file and load dynamic configuratinons from (it is being checked periodically)"
+        },
+    )
+    dynamic_config_timeout: float = field(
+        default=30.0, metadata={"help": "dynamic configuration state timeout"}
     )
 
 
@@ -737,3 +749,49 @@ class MetaseqConfig(MetaseqDataclass):
     lr_scheduler: Any = MISSING
     bpe: Any = MISSING
     tokenizer: Any = None
+
+
+class DynamicConfig(dict):
+    """
+    can be used instead of redis to store updatable key-value
+
+    test:
+        data = {"a": 1}
+        with open('data.json', 'w') as fp:
+            json.dump(data, fp)
+        dcfg = DynoCfg(json_file_path = "data.json", timeout = 1)
+        print(dcfg["a"]) # > 1
+        data = {"a": 4}
+        with open('data.json', 'w') as fp:
+            json.dump(data, fp)
+        print(dcfg["a"]) # > 1
+        time.sleep(1)
+        print(dcfg["a"]) # > 4
+    """
+
+    default_state = {"force_profile": False}
+
+    def __init__(self, json_file_path=None, timeout=30):  # timeout in sec
+        self.timeout = timeout
+        self.timer_start = time.time()
+        self.json_file_path = json_file_path
+        if self.json_file_path is not None:
+            with open(self.json_file_path, "r") as json_file:
+                super().__init__(**json.load(json_file))
+        else:
+            super().__init__(**DynamicConfig.default_state)
+
+    def refresh(self):
+        if self.json_file_path is not None:
+            # if data have not been updated in a while, reload from the file
+            # can speed up if were to calculate hash sum of the file to monitor updates
+            timer_value = time.time() - self.timer_start
+            if timer_value > self.timeout:
+                with open(self.json_file_path, "r") as json_file:
+                    refreshed_config = json.load(json_file)
+                    super().__init__(**refreshed_config)
+                self.timer_start = time.time()
+
+    def __getitem__(self, key):
+        self.refresh()
+        return super().__getitem__(key)


metaseq/cli/train.py
```diff
@@ -30,19 +30,28 @@
...
 
-logging.basicConfig(
-    format="%(asctime)s | %(levelname)s | %(name)s | %(message)s",
-    datefmt="%Y-%m-%d %H:%M:%S",
-    level=os.environ.get("LOGLEVEL", "INFO").upper(),
-    stream=sys.stdout,
-)
+if "SLURM_PROCID" in os.environ:
+    logging.basicConfig(
+        format=f"slurm_procid {os.environ['SLURM_PROCID']} : %(asctime)s | %(levelname)s | %(name)s | %(message)s",
```
dgrnbrg-meta, line 43: Rather that duplicating the log config, could we just include a string that might be `''`?

metaseq/cli/train.py
```diff
@@ -308,9 +317,18 @@ def train(
...
+        json_file_path=cfg.common.dynamic_config_path,
+        timeout=cfg.common.dynamic_config_timeout,
+    )
+
     for i, samples in enumerate(progress):
-        if distributed_utils.get_global_rank() == 0 and cfg.common.profile and i == 5:
-            logger.info("STARTING PROFILER")
+        force_profile = dcfg["force_profile"]
+        do_profile = (distributed_utils.get_global_rank() == 0) and (
+            (cfg.common.profile and i == 5) or force_profile
```
dgrnbrg-meta, line 328: Why `i == 5`?

Should `cfg.common.profile` still be a way to enable tracing, or should we just switch to the dynamic configs?

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
     lr_scheduler: Any = MISSING
     bpe: Any = MISSING
     tokenizer: Any = None
+
+
+class DynamicConfig(dict):
```
dgrnbrg-meta, line 753: Should this extend `dict`? I would think that just implementing `__getitem__` might make sense, because setting items would lead to weird states.

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+        print(dcfg["a"]) # > 4
+    """
+
+    default_state = {"force_profile": False}
+
+    def __init__(self, json_file_path=None, timeout=30):  # timeout in sec
+        self.timeout = timeout
+        self.timer_start = time.time()
+        self.json_file_path = json_file_path
+        if self.json_file_path is not None:
```
dgrnbrg-meta, line 777: Does this make sense as a use case? Or should there always be a config path required?

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+            with open(self.json_file_path, "r") as json_file:
+                super().__init__(**json.load(json_file))
```
dgrnbrg-meta, line 779: It would be good to only implement the loading code once. Could this call `self.refresh()`?

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+        else:
+            super().__init__(**DynamicConfig.default_state)
+
+    def refresh(self):
+        if self.json_file_path is not None:
+            # if data have not been updated in a while, reload from the file
+            # can speed up if were to calculate hash sum of the file to monitor updates
+            timer_value = time.time() - self.timer_start
+            if timer_value > self.timeout:
+                with open(self.json_file_path, "r") as json_file:
```
dgrnbrg-meta, line 789: This could throw an exception if the file doesn't exist

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+                    refreshed_config = json.load(json_file)
```
dgrnbrg-meta, line 790: This could throw an exception if the file is partially written/isn't valid JSON

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+                    super().__init__(**refreshed_config)
```
dgrnbrg-meta, line 791: Is it safe to re-init objects?

metaseq/dataclass/configs.py
```diff
@@ -737,3 +748,49 @@ class MetaseqConfig(MetaseqDataclass):
...
+        self.timeout = timeout
+        self.timer_start = time.time()
+        self.json_file_path = json_file_path
+        if self.json_file_path is not None:
+            with open(self.json_file_path, "r") as json_file:
+                super().__init__(**json.load(json_file))
+        else:
+            super().__init__(**DynamicConfig.default_state)
+
+    def refresh(self):
```
dgrnbrg-meta, line 783: Is there any situation in which we'd need this to be thread safe?

metaseq/cli/train.py
```diff
@@ -308,9 +317,18 @@ def train(
...
 
+    dcfg = DynamicConfig(
+        json_file_path=cfg.common.dynamic_config_path,
+        timeout=cfg.common.dynamic_config_timeout,
+    )
+
     for i, samples in enumerate(progress):
-        if distributed_utils.get_global_rank() == 0 and cfg.common.profile and i == 5:
-            logger.info("STARTING PROFILER")
+        force_profile = dcfg["force_profile"]
```
dgrnbrg-meta, line 326: Should `force_profile` cover all ranks, or should the ranks be selectable?

@anupambhatnagar

dgrnbrg-meta (dismissed):
This looks like a great feature that'll be really useful!

I was wondering: do you think that the schema should be validated? I am imagining someone manually changing the json file, and making a typo in a field so that it's valid, parseable json; but it's not a key that we consider.

metaseq/cli/train.py
igormolybogFB: I am trying to make this backward-compatible so that I don't break anyone else's classical way of doing things. Ideally, all of the logging/profiling work should be controlled by dynamic configuration, but that might be unusual for the rest of the team if included in the very first PR

metaseq/dataclass/configs.py
igormolybogFB: Done purely for backward-compatibility and in case using dynamic configs impacts performance and needs to be turned off.

metaseq/dataclass/configs.py
igormolybogFB: Did not have problems with this when testing, but I think I will change this when issue the next patch

metaseq/dataclass/configs.py
igormolybogFB: What do you mean thread safe? Like if this code is executed in multiple thread at once? that is not intended but I also don't see how this would break if it does

metaseq/cli/train.py
dgrnbrg-meta: I like that this doesn't break others' way of doing things.

I am still concerned about `i == 5` :)

igormolybogFB: i == 5 is an arbitrary choice of @ngoyal2707

metaseq/dataclass/configs.py
dgrnbrg-meta: > Like if this code is executed in multiple thread at once? 

Exactly that. The issue would be that `self.timer_start` isn't managed atomically. It seems like the worst case would be that the file is read multiple times, so it's probably fine to ignore.

metaseq/dataclass/configs.py
```diff
@@ -737,3 +749,56 @@ class MetaseqConfig(MetaseqDataclass):
...
+                except json.JSONDecodeError as jsonerror:
+                    logging.warning(
+                        f"""Error refreshing dynamic config: reading file {json_file}
+                    resulted in JSONDecodeError {jsonerror}"""
+                    )
+                except IOError as ioerror:
+                    logging.warning(
+                        f"""Error refreshing dynamic config: reading file {json_file}
+                    resulted in IOError {ioerror}"""
+                    )
```
dgrnbrg-meta, line 800: Could you add 3 tests for this?

1. Test that a malformed (truncated) file catches the error.
2. Test that an empty file catches the error.
3. Test that a file that doesn't exist catches the error.

My concern is that, with my limited knowledge of python exception types, a different exception type could be generated in a rare situation and impact the big training at an unexpected/inopportune time.

metaseq/dataclass/configs.py
igormolybogFB: It is intended that the files are read as many times as there are processes or threads - this way every worker gets updated dictionary. I still don't see problems here

dgrnbrg-meta:
> @dgrnbrg-meta Regarding schema validation: I think this would be a cool feature but I also think that this is not a blocker for profiling and analyzing hardware issues

I don't think that schema validation needs to be fancy (e.g. with a schema validation API). Instead, something like this would go far for usability:

```python
valid_keys = ["force_profile"]
for k,v in self.data.items():
  if k not in valid_keys:
    logging.warning(f"Encounterd unknown dynamic config option: {k} {v}")
```

With that, users would have clear feedback about typos.

anupambhatnagar:
@igormolybogFB In `train.py` you are enabling the trace collection with the python call stack on line 329. I strongly recommend testing this out on a large model (6.7B) before merging the changes into main. The pytorch profiler has had issues that were patched about three weeks ago so the success of this change depends on using the correct pytorch version along with your code change.

igormolybogFB:
@anupambhatnagar line 329 is not changed in this PR. I agree that large-scale stack tracing must be tested and I have that in my backlog. This PR is not about testing stack_trace but about dynamic configs.

anupambhatnagar:
> @anupambhatnagar line 329 is not changed in this PR. I agree that large-scale stack tracing must be tested and I have that in my backlog. This PR is not about testing stack_trace but about dynamic configs.

You are correct that this PR doesn't change line 329 but this PR would lead to that piece of code being invoked more often than before hence I suggest being cautious about this change.

suchenzang (approved):
Test on our new compute env and see! 😅

PR #474, opened 2022-11-01
Resume on requeue

sriniiyer:
Resume on requeue fails right now because we aren't saving checkpoint_last but instead, are saving cherkpoints per interval, like checkpoint_1_600.pt. Commenting this allows resuming even though checkpoint_last isnt saved.

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -314,11 +314,13 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
             first_launch = True
             checkpoint_path_to_load = cfg.restore_file
 
-    if cfg.restore_file != default_restore_file and cfg.finetune_from_model:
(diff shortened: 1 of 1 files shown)


todpole3 (commented):
It seems that the original check was there for a purpose -- to verify fine-tuning starts from the correct checkpoint.

Is there a way to implement resuming fine-tuning without removing this original check?

sriniiyer:
> It seems that the original check was there for a purpose -- to verify fine-tuning starts from the correct checkpoint.
> 
> Is there a way to implement resuming fine-tuning without removing this original check?

The original check is broken, it crashes on requeue. This version does resume from the correct checkpoint - I tested it by saving every 200 steps and requeued after 650 steps, and it continued from the correct checkpoint. 

Is there a use case you know of where this fails?

todpole3 (approved):
Approve to unblock

Issue #477, opened 2022-11-01
PathManager has no attribute `supports_rename`


andrewPoulton:
## 🐛 Bug

Following #441, a regression occurs when running `metaseq/launcher/opt_baselines.py`, due to the call to `PathManager.supports_rename` on L536 in `checkpoint_utils.py`, but this attribute no longer exists after #441 was merged.

### To Reproduce

Run (eg from @Xirider)
```
python metaseq/launcher/opt_baselines.py --prefix train.8m --model-size 8m --checkpoints-dir ./test-checkpoint --tensorboard-logdir ./test-checkpoint --num-trials 1 --azure --num-gpus 4 --num-nodes 1 --seed 1 --circleci --local --disable-validation --max-epoch 100 --max-update 100
```

davides:
Closing as the fix was merged in https://github.com/facebookresearch/metaseq/pull/478

PR #441, opened 2022-10-21
Remove dependency on iopath.

davides:
**Patch Description**
This change decouples metaseq and iopath with the following:
* Copy the main classes into the local file_io package
* Copy the associated unit tests
* Drop functionality not needed in metaseq: TabularIO, telemetry
* Update setup.py. `portalocker` was previously a transitive dependency, and we still need it for `metaseq.file_io.common.file_lock`
* Pull forward the pending AzureBlobPathHandler from https://github.com/facebookresearch/iopath/pull/17
  * Some controls have been put in place so that read/write operations use a known amount of memory when dealing with larger files:
    * `_open("wb", buffering=<buffer-size>)` will buffer up to the requested amount of data in memory before flushing it to the service with the PutBlock operation
    * `_open("rb", buffering=<buffer-size>)` will use the Blob client's chunk iterator to only download a fixed amount of data at a time
    * `_close()` in write-mode will flush any buffered data with one more PutBlock, and finalize the blob with PutBlockList
    * The block-based approach should work for both block blobs and append blobs (see [the Azure docs](https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs)).

**Testing steps**
```
$ python -m unittest discover tests/file_io/
.10/21/2022 12:23:17 PM Caching az://lrsstoragewest3/data/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file1.bin ...
10/21/2022 12:23:17 PM URL az://lrsstoragewest3/data/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file1.bin cached in /tmp/tmpy3w003kj/blob_cache/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file1.bin
10/21/2022 12:23:17 PM Caching az://lrsstoragewest3/data/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file2.bin ...
10/21/2022 12:23:17 PM URL az://lrsstoragewest3/data/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file2.bin cached in /tmp/tmpy3w003kj/blob_cache/temp/unittest/68b4fa21-491a-40fc-b048-9d18d528d00b/testdir/file2.bin
[...]

PR #478, opened 2022-11-01
Re-add supports_rename and rename to PathManager

andrewPoulton:
Addresses #477, and also adds `rename` method to PathManager which is also called in `checkpoint_utils.torch_persistent_save`.

Currently untested, @Xirider should run in his env, as he found the bug.

### DIFF
--- a/metaseq/file_io/common/__init__.py
+++ b/metaseq/file_io/common/__init__.py
@@ -741,6 +741,14 @@ class PathManager:
                 return True
         return False
 
+    def supports_rename(self, path: str) -> bool:
+        # PathManager doesn't yet support renames
+        return not self.path_requires_pathmanager(path)
+
+    @staticmethod
+    def rename(src: str, dst: str):
+        os.rename(src, dst)
+
     # pyre-fixme[24]: Generic type `os.PathLike` expects 1 type parameter.
     def __get_path_handler(self, path: Union[str, os.PathLike]) -> PathHandler:
         """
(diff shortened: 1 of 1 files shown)


metaseq/file_io/common/__init__.py
```diff
@@ -741,6 +741,14 @@ def path_requires_pathmanager(self, path: str) -> bool:
                 return True
         return False
 
+    def supports_rename(self, path: str) -> bool:
+        # PathManager doesn't yet support renames
+        return not self.path_requires_pathmanager(path)
+
+    @staticmethod
+    def rename(src: str, dst: str):
```
davides, line 749: I believe this should be an instance method. `metaseq.file_io.PathManager` was previously a static wrapper around iopath's global instance, but I changed it to be a global instance as well as part of https://github.com/facebookresearch/metaseq/pull/441.

You should also be able to confirm these are working as expected by adding new/updated tests [here](https://github.com/facebookresearch/metaseq/blob/main/tests/file_io/test_native.py) and running `python -m tests.file_io.test_native`

Xirider:
Tested it in both of my envs and everything works!

Xirider (approved):

metaseq/file_io/common/__init__.py
suchenzang: +1; might be even safer to call `supports_rename` from within `rename` too.

andrewPoulton: Gotcha @davides, thanks. Also added the `supports_rename` guard in `rename` (raising a `ValueError` if it doesn't support rename - didn't want it silently fail. The only place `rename` is called in the codebase is immediately following a `supports_rename` check anyway, so this won't break anything elsewhere).

PR #479, opened 2022-11-02
isolated PR for seq_par with autograd fn to have more control over op…

ngoyal2707:
custom autograd function from transformer block to have more control over which ops to recompute.
I know the code is ugly and adds another codepath for model :( 

I think maybe we can just kill model parallel modules at some point?

### DIFF
--- a/metaseq/model_parallel/models/transformer_lm.py
+++ b/metaseq/model_parallel/models/transformer_lm.py
@@ -52,6 +52,18 @@ class ModelParallelTransformerLanguageModel(TransformerLanguageModel):
             args, "use_sharded_state", False
         ), "Use sharded state must be True for tensor parallel, otherwise model saving and loaded might be broken"
 
+        if getattr(args, "sequence_parallel", False):
+            assert (
+                getattr(args, "model_parallel_size", 1) > 1
+            ), "--sequence-parallel only works when --model-parallel-size is greater than 1"
+            assert (
+                getattr(args, "dropout", 0.0) == 0.0
+            ), "havent yet tested if rng states are correct for dropout with seq_parallel"
+            assert (
+                getattr(args, "activation_fn", 'gelu') == 'gelu'
+            ), "For now only supports gelu"
+
+
         decoder = ModelParallelTransformerDecoder(
             args,
             task.target_dictionary,
--- a/metaseq/model_parallel/models/transformer_lm.py
+++ b/metaseq/model_parallel/models/transformer_lm.py
@@ -60,9 +60,8 @@ class ModelParallelTransformerLanguageModel(TransformerLanguageModel):
                 getattr(args, "dropout", 0.0) == 0.0
             ), "havent yet tested if rng states are correct for dropout with seq_parallel"
             assert (
-                getattr(args, "activation_fn", 'gelu') == 'gelu'
+                getattr(args, "activation_fn", "gelu") == "gelu"
             ), "For now only supports gelu"
-
 
         decoder = ModelParallelTransformerDecoder(
             args,
--- a/metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
+++ b/metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
@@ -1,0 +1,299 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+import torch
+
+import fused_layer_norm_cuda
+
+from metaseq.modules.activation_functions import gelu, gelu_back
+
+try:
+    from megatron.mpu.mappings import (
+        _reduce_scatter_along_first_dim,
+        _gather_along_first_dim
+    )
+    from megatron.mpu.utils import split_tensor_along_last_dim
+    from megatron.model.fused_softmax import scaled_upper_triang_masked_softmax_cuda
+    has_megatron_submodule = True
+except (ImportError, ModuleNotFoundError):
+    has_megatron_submodule = False
+
+
+
+class SequeuceParallelTransformerBlock(torch.autograd.Function):
+    """
+    This is custom FFN autograd function hardcoded for:
+    bias: false,
+    layernorm affine: false, ln eps: 1e-5
+    sequence_parallel: true,
+    activation: gelu,
+    gelu, layernorm: always recomputed i.e. no activation memory for these
+    """
+    @staticmethod
+    def forward_mha(q, k, v, bsz, seq_len, head_dim, embed_dim_per_partition, dtype):
+        scaling = head_dim ** -0.5
+        matmul_result = torch.empty(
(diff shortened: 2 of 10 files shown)


metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
```diff
@@ -0,0 +1,299 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+import torch
+
+import fused_layer_norm_cuda
```
suchenzang, line 9: Is this coming from apex installation?  Test are failing to import this...

metaseq/model_parallel/models/transformer_lm.py
```diff
@@ -52,6 +52,18 @@ def build_model(cls, args, task):
...
+        if getattr(args, "sequence_parallel", False):
+            assert (
+                getattr(args, "model_parallel_size", 1) > 1
+            ), "--sequence-parallel only works when --model-parallel-size is greater than 1"
+            assert (
+                getattr(args, "dropout", 0.0) == 0.0
+            ), "havent yet tested if rng states are correct for dropout with seq_parallel"
+            assert (
+                getattr(args, "activation_fn", 'gelu') == 'gelu'
+            ), "For now only supports gelu"
```
suchenzang, line 64: Will there be a follow-up PR to add for swiglu? 😅

metaseq/modules/transformer_decoder_layer.py
```diff
@@ -199,6 +200,18 @@ def forward(
         Returns:
             encoded output of shape `(seq_len, batch, embed_dim)`
         """
+        if getattr(self.args, "sequence_parallel", False):
```
suchenzang, line 203: To confirm I'm following...
* We are still calling into ModelParallelTransformerDecoderLayer's build_fc1, build_fc2, etc. methods.
* These components (fc1, fc2, etc.) are then passed into `SequeuceParallelTransformerBlock`
* Killing model parallel modules going forward would imply that all models will be trained with sequence parallel

Separately, have a nit about naming since we separately have `TransformerDecoderMultiLayerBlockModule` floating around - is that still needed? If not, can open a task to remove (assume fsdp wrap every layer?) and keep the current `SequeuceParallelTransformerBlock` naming.

metaseq/models/transformer_decoder.py
```diff
@@ -422,6 +422,7 @@ def extract_features(
                 incremental_state=incremental_state,
                 self_attn_mask=self_attn_mask,
                 self_attn_padding_mask=self_attn_padding_mask,
+                recompute_fc1=(idx < getattr(self.args, "recompute_fc1_num_layers", 0)),
```
suchenzang, line 425: Is this.. the first N layers by design? Sorry, don't have much context here.

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
```diff
@@ -0,0 +1,299 @@
...
+        )
+        grad_k = grad_k.transpose(0, 1).contiguous().view(seq_len, bsz, -1)
+        grad_kvq_proj_output = torch.cat([grad_k, grad_v, grad_q], dim=-1)
+        return grad_kvq_proj_output
+
+    @staticmethod
+    def forward(
+        ctx,
+        input,
+        kvq_proj_weight,
```
suchenzang, line 87: I think I might've missed a transpose that happened somewhere..? `self.self_attn.qkv_proj.weight` is passed in here and it seems like this is intentionally kvq for a reason...

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
```diff
@@ -0,0 +1,299 @@
...
+        q  = q.view(seq_len, -1, head_dim)
+        k  = k.view(seq_len, -1, head_dim)
+        v  = v.view(seq_len, -1, head_dim).transpose(0, 1)
+
+        attn, _ = SequeuceParallelTransformerBlock.forward_mha(q, k, v, bsz, seq_len, head_dim, embed_dim_per_partition, dtype)
+
+        out_proj_out = torch.matmul(attn, out_proj_weight.t())
+        out_proj_out = _reduce_scatter_along_first_dim(out_proj_out)
+
+        # out_proj_out = out_proj_out + residual
```
suchenzang, line 127: Nit: remove?

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
```diff
@@ -0,0 +1,299 @@
...
+        out_proj_out = out_proj_out + residual
+
+        # Take out residual connection for FFN
```
suchenzang, line 130: Nit: `out_proj_out` seems to contain residual at this point, which gets passed into FFN... how residual get "taken out" here?

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
suchenzang: Ah this looks like just a naming thing? Previously we had `kvq, _ = self.qkv_proj(query)`, which is just Line 113 below

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
suchenzang: Oh maybe `out_proj_out = out_proj_out + residual` is supposed to be removed in Line 128?

suchenzang: Oh nvm, taken out == saved

metaseq/modules/transformer_decoder_layer.py
ngoyal2707: no we need to kill "TransformerDecoderMultiLayerBlockModule", I can send a separate PR for that

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
ngoyal2707: makes sense, moved the comment up

metaseq/models/transformer_decoder.py
ngoyal2707: will add comment about it

suchenzang (approved):
Approving to unblock, assuming some of the tests eventually pass and... hoping that all the linear algebra matrix derivative transposes are just right 😅

metaseq/model_parallel/modules/sequence_parallel_transformer_layer.py
stephenroller: hm, why isn't it megatron.fused_kernels.fused_layer_norm?

ngoyal2707: sorry its coming from apex, https://github.com/NVIDIA/apex/blob/master/apex/normalization/fused_layer_norm.py#L36-L37

PR #480, opened 2022-11-02
Copy to/from NFS with subdirectories.

zdevito:
If you set cloud_upload_path to "nfs:<some_local_directory>" this will save the checkpoints to local disk as normal, but then copy to this directory similar to how we do azcopy, except to a local folder mounted in nfs. To avoid many files in the same directory, it adds a directory for each checkpoint. It uses these directories to discover the most recent checkpoint on restart.

Directories are initially created with a leading _ and only renamed when the checkpoint successfully completes.

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -10,7 +10,6 @@ import logging
 import os
 import re
 import traceback
-from glob import glob
 from typing import Any, Dict, List, Optional, Tuple
 import shutil
 
@@ -246,6 +245,10 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
             if restart_from_latest:
                 max_checkpoint = None
                 for candidate in os.listdir(nfs_path):
+                    if candidate == "checkpoint_last":
+                        raise RuntimeError(
+                            "trying to restart a job that already wrote checkpoint_last"
+                        )
                     m = re.match(r"checkpoint_(\d+)", candidate)
                     if m and (max_checkpoint is None or int(m[1]) > max_checkpoint):
                         max_checkpoint = int(m[1])
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -10,7 +10,9 @@ import logging
 import os
 import re
 import traceback
+from glob import glob
 from typing import Any, Dict, List, Optional, Tuple
+import shutil
 
 import torch
 from omegaconf import OmegaConf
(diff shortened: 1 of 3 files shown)


metaseq/checkpoint_utils.py
```diff
@@ -229,42 +230,62 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
...
                 cfg.save_dir, "checkpoint_last{}.pt".format(suffix)
             )
-            azure_utils.download_specific_ckpt(blob_url, checkpoint_path_to_load)
-        else:
-            logger.info(
-                f"Using checkpoint {checkpoint_path_to_load} even while on Azure"
-            )
+            nfs_path = cfg.cloud_upload_path[4:]
+            filename = None
+            if cfg.restore_file == default_restore_file or int(os.environ.get("SLURM_RESTART_COUNT", 0)) > 0:
```
ruanslv, line 241: Could we encapsulate the three checks in a separate method? We should also check for `finetune_from_model` here too, no?

`if (
                # --restore-file was not passed, always download latest checkpoint
                (
                    cfg.restore_file == default_restore_file
                    and cfg.finetune_from_model is None
                )
                # --restore-file was passed, but we requeued, so download latest checkpoint
                or int(os.environ.get("SLURM_RESTART_COUNT", 0)) > 0
            ):`

metaseq/checkpoint_utils.py
```diff
@@ -229,42 +230,62 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
...
+                max_checkpoint = None
+                for candidate in os.listdir(nfs_path):
+                    m = re.match(r"checkpoint_(\d+)", candidate)
+                    if m and (max_checkpoint is None or int(m[1]) > max_checkpoint):
```
ruanslv, line 245: We also have a checkpoint called `last` at the end of the run, that wouldn't have number of updates.

I guess this depends on how we implement the save_checkpoint logic in NFS.

In general code looks good but worth testing, should we implement save_checkpoint logic first? (I got confused with the PR title, this is copying from NFS to SSD, not copying to NFS)

zdevito: So I've seen `_last` as a locally named file, but it doesn't seem to to make it to azure. See `_choose_last_completely_uploaded_ckpt` in metaseq_internal where it is just looking at a `checkpoint_\d` pattern.

metaseq/cli/train.py
```diff
@@ -455,6 +463,15 @@ def post_checkpoint_callback(cfg, filename):
                 f"Successfully copied {filename} to {cfg.checkpoint.cloud_upload_path}"
             )
             os.remove(filename)
+        elif cfg.checkpoint.cloud_upload_path.startswith("nfs:"):
```
zdevito, line 466: This is `save_checkpoint` logic.

metaseq/cli/train.py
```diff
@@ -418,7 +421,7 @@ def validate_and_save(
 
     should_stop |= should_stop_early(cfg, valid_losses[0])
 
-    # Save checkpoint
+    # Save checkpointC
```
stephenroller, line 424:
```suggestion
    # Save checkpoint
```

metaseq/cli/train.py
```diff
@@ -470,6 +480,34 @@ def post_checkpoint_callback(cfg, do_evaluate, eval_kwargs, filename):
...
+                    destination_checkpoints_dir,
+                    temporary_checkpoint_dir,
+                    checkpoint_file,
+                ),
+            )
+            torch.distributed.monitored_barrier(
+                group=eval_kwargs["gloo_pg"], timeout=timedelta(minutes=5)
+            )
+            if distributed_utils.get_global_rank() == 0:
+                os.rename(
```
stephenroller, line 506:
```suggestion
                # atomic rename of the final checkpoint directory, now that all workers have completed
                # their copies
                os.rename(
```

metaseq/cli/train.py
```diff
@@ -470,6 +480,34 @@ def post_checkpoint_callback(cfg, do_evaluate, eval_kwargs, filename):
...
+            checkpoint_dir, checkpoint_file = _checkpoint_add_directory(basename)
+            destination_checkpoints_dir = cfg.checkpoint.cloud_upload_path[4:]
+            temporary_checkpoint_dir = f"_{checkpoint_dir}"
+            try:
+                os.mkdir(
+                    os.path.join(destination_checkpoints_dir, temporary_checkpoint_dir)
+                )
+            except FileExistsError:
+                pass  # another worker got here first
+            shutil.copyfile(
```
stephenroller, line 494:
```suggestion
            # copy the checkpoint from local storage to nfs in the background
            shutil.copyfile(
```

stephenroller (approved):
i think this looks good! thanks!

metaseq/checkpoint_utils.py
ruanslv: `_choose_last_completely_uploaded_ckpt` is broken, we do have last checkpoints

metaseq/checkpoint_utils.py
```diff
@@ -228,42 +230,71 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
...
+        elif cfg.cluster_env == ComputeEnvs.AZURE.value and has_metaseq_internal:
+            if (
+                # --restore-file was not passed, always download latest checkpoint
+                (
+                    cfg.restore_file == default_restore_file
+                    and cfg.finetune_from_model is None
+                )
+                # --restore-file was passed, but we requeued, so download latest checkpoint
+                or int(os.environ.get("SLURM_RESTART_COUNT", 0)) > 0
+            ):
```
ruanslv, line 274: `if restart_from_latest:`

metaseq/checkpoint_utils.py
```diff
@@ -228,42 +230,71 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
...
+                # download checkpoint into local save_dir
+                checkpoint_path_to_load = os.path.join(
+                    cfg.save_dir, "checkpoint_last{}.pt".format(suffix)
+                )
+                azure_utils.download_recent_ckpt(
+                    cfg.cloud_upload_path, checkpoint_path_to_load, suffix + ".pt"
+                )
+            elif (
+                # --restore-file was passed and is a blob URL, download that checkpoint
+                cfg.restore_file != default_restore_file
```
ruanslv, line 284: replace by `specific_restore_file_provided`

metaseq/checkpoint_utils.py
```diff
@@ -228,42 +230,71 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
...
+                )
+                azure_utils.download_specific_ckpt(blob_url, checkpoint_path_to_load)
+            else:
+                logger.info(
+                    f"Using checkpoint {checkpoint_path_to_load} even while on Azure"
+                )
+
     # RSC logic: --restore-file was passed, and we requeued
     elif (
         cfg.restore_file != default_restore_file
```
ruanslv, line 300: replace by `specific_restore_file_provided`

metaseq/cli/train.py
```diff
@@ -470,6 +480,37 @@ def post_checkpoint_callback(cfg, do_evaluate, eval_kwargs, filename):
...
             )
             os.remove(filename)
+        elif cfg.checkpoint.cloud_upload_path.startswith("nfs:"):
+            path, basename = os.path.split(filename)
+            checkpoint_dir, checkpoint_file = _checkpoint_add_directory(basename)
+            destination_checkpoints_dir = cfg.checkpoint.cloud_upload_path[4:]
+            temporary_checkpoint_dir = f"_{checkpoint_dir}"
+            try:
+                os.mkdir(
+                    os.path.join(destination_checkpoints_dir, temporary_checkpoint_dir)
```
tangbinh, line 490: Should we be using `makedirs` to create the directories recursively if needed? It might also be a bit cleaner if we use `exist_ok=True` to avoid the race condition.

zdevito: I used mkdir so it wouldn't do something surprising if intermediate directories didn't already exist. The exception logic is just what `makedirs` would do for a single dir anyway.

PR #488, opened 2022-11-04
Add data proportions

sriniiyer:
### PR to control data proportions

PER BENCHMARK PROPORTIONS

--data-sampling-prob '{"flan": 0.35, "ni_v2": 0.24, "exmix": 0.03, "crossfit": 0.03, "pretrain": 0.0, "cot": 0.02, "t5": 0.03, "promptsource": 0.28, "unified_skg": 0.02,}'

- You don't need to specify probs for all benchmarks. The code will uniformly distribute the remaining prob mass amongst the rest of the benchmarks. So, you can specify just --data-sampling-prob '{"pretrain": 0.0}'

- The logs print the before proportions, and the after proportions, which you can double check.

EQUALIZE CLUSTER PROPORTIONS AFTER APPLYING BENCHMARK PROPS

--equalize-cluster-probs

This will equalize the prob of each cluster within a benchmark. Note that this needs cluster names to be included in the dataset names.

PROVIDE PER BENCHMARK EPS

--caps '{"flan": 30000, "ni_v2": 5000, "exmix": 20000, "crossfit": 20000, "cot": 10000, "t5": 20000, "promptsource": 20000, "unified_skg": 20000,}'

Instead of applying a uniform eps, this helps provide a per benchmark eps. If its not provided for a benchmark, it will fall back to the default eps, and if no default is specified, it will fall back to the dataset length.

### DIFF
--- a/metaseq/tasks/streaming_language_modeling.py
+++ b/metaseq/tasks/streaming_language_modeling.py
@@ -95,6 +95,10 @@ class StreamingLanguageModelingConfig(MetaseqDataclass):
         default='{}',
         metadata={"help": "Proportion of each finetuning benchmark to use. Use only during finetuning"}
     )
+    caps: Optional[str] = field(
+        default='{}',
+        metadata={"help": "Proportion of each finetuning benchmark to use. Use only during finetuning"}
+    )
     data_subshard_count: int = field(
         default=1,
         metadata={
@@ -193,19 +197,14 @@ class StreamingLanguageModelingTask(LegacyTask):
             + [self.eod]
         )
 
-    def _get_sample_prob(self, dataset_lens):
+    def _get_sample_prob(self, dataset_caps):
         """
         Get smoothed sampling porbability by corpus. This helps small corpus by upsampling them.
         """
-        if self.args.multicorpus_sampling_maximum == DEFAULT_MULTICORPUS_MAX:
-            prob = dataset_lens / dataset_lens.sum()
-            smoothed_prob = prob**self.args.multicorpus_sampling_alpha
-            smoothed_prob = smoothed_prob / smoothed_prob.sum()
-        else:
-            dataset_lens = np.array(
-                [min(l, self.args.multicorpus_sampling_maximum) for l in dataset_lens]
-            )
-            smoothed_prob = dataset_lens / sum(dataset_lens)
+        prob = dataset_caps / sum(dataset_caps)
+
+        smoothed_prob = prob**self.args.multicorpus_sampling_alpha
+        smoothed_prob = smoothed_prob / smoothed_prob.sum()
         return smoothed_prob
 
     def _alpha_sampling(self, datasets, corpora, epoch=1):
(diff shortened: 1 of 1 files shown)


metaseq/tasks/streaming_language_modeling.py
```diff
@@ -211,24 +212,41 @@ def _alpha_sampling(self, datasets, corpora, epoch=1):
...
+        for i, s in enumerate(corpora):
+            benchmark = s.split('__')[0]
+            sample_probs[i] = sample_probs[i] * data_sampling_prob[benchmark] / per_benchmark_prob[benchmark]
+
+        new_per_benchmark_prob = {}
+        for i, s in enumerate(corpora):
+            benchmark = s.split('__')[0]
+            if benchmark not in new_per_benchmark_prob:
+                new_per_benchmark_prob[benchmark] = 0.0
+            new_per_benchmark_prob[benchmark] += sample_probs[i]
```
ramakanth-pasunuru, line 246: Can we add some more assertions like new_per_benchmark_prob matches the intended proportions given in data_sampling_prob?

ramakanth-pasunuru (commented):
Let me know once you add the cluster balancing as well.

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -213,26 +216,60 @@ def _alpha_sampling(self, datasets, corpora, epoch=1):
...
+        # scale up/down the remaining proportions uniformly
+        total_benchmarks = len(per_benchmark_prob)
+        total_assigned_prob = sum(data_sampling_prob.values())
+        total_assigned_benchmarks = len(data_sampling_prob)
+
+        # Assign probs for the benchmarks not present in the dictionary
+        for benchmark in per_benchmark_prob:
+            if benchmark not in data_sampling_prob:
+                benchmark_prob = per_benchmark_prob[benchmark] if not self.args.equalize_cluster_probs else sum(per_benchmark_prob[benchmark].values())
+                data_sampling_prob[benchmark] = benchmark_prob * (1.0 - total_assigned_prob)
```
ramakanth-pasunuru, line 252: Previously with pre-train data proportion, we use to keep the task eps but not balance the benchmarks. But, now it seems like that case is not possible given that we always rebalance the benchmarks when only pre-train proportion is mentioned in data_sampling_prob arg. I think that is fine, but just want to make sure that's the case.

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -213,26 +216,60 @@ def _alpha_sampling(self, datasets, corpora, epoch=1):
...
+
+        # Scale prob for tasks in each benchmark
+        for i, s in enumerate(corpora):
+            benchmark = s.split('__')[0]
+            current_benchmark_prob = sum(per_benchmark_prob[benchmark].values()) if self.args.equalize_cluster_probs else per_benchmark_prob[benchmark]
+            sample_probs[i] = sample_probs[i] * data_sampling_prob[benchmark] / current_benchmark_prob
```
ramakanth-pasunuru, line 258: I guess we cannot set cluster balancing with out balancing the benchmarks first right?

metaseq/checkpoint_utils.py
```diff
@@ -179,18 +179,19 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
-def verify_shards(cfg, dir=None, checkpoint_name=None):
+def verify_shards(cfg, dir=None, checkpoint_name=None, distributed_world_size=None):
```
ramakanth-pasunuru, line 182: This seems to be from another PR. May be remove the changes from this file

metaseq/tasks/streaming_language_modeling.py
sriniiyer: correct

metaseq/checkpoint_utils.py
sriniiyer: done

ramakanth-pasunuru (approved):
LGTM

PR #491, opened 2022-11-04
Replace jobid in local checkpoint dir with actual job id 

suchenzang:


stephenroller (approved):
I think it's the only way

PR #495, opened 2022-11-07
Speech demo

lilisierrayu:
**Patch Description**
Making the inference task tokenizer match with cm3_streaming. Note that the inference task is very redundant. Zetta team is working on unifying them, will adopt their change once they did that. 

**Testing steps**

### DIFF
--- a/metaseq/service/constants.py
+++ b/metaseq/service/constants.py
@@ -9,30 +9,27 @@ MAX_SEQ_LEN = 2048
 BATCH_SIZE = 2048  # silly high bc we dynamically batch by MAX_BATCH_TOKENS
 MAX_BATCH_TOKENS = 3072
 DEFAULT_PORT = 6010
-MODEL_PARALLEL = 4
-TOTAL_WORLD_SIZE = 4
+MODEL_PARALLEL = 2
+TOTAL_WORLD_SIZE = 2
+MAX_BEAM = 32
 
-
-CHECKPOINT_FOLDER = "/data/gpt-z/cm3/models/ablations/causal_one_image/resharded"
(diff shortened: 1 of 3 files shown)


metaseq/hub_utils.py
```diff
@@ -320,13 +320,13 @@ def generate(
                     tokens_no_eos = tokens[1:] if echo else tokens
                     scores_with_eos = [None] + scores[1:] if echo else scores
                     # turn it into a string
-                    text = self.bpe.decode(tokens_no_eos)
+                    text = self.bpe.bpe.decode(tokens_no_eos)
```
ArmenAg, line 323: this wont call the hf_byte_bpe decode, in our case we need to set skip_special_tokens to False to generate speech tokens.

metaseq/hub_utils.py
```diff
@@ -320,13 +320,13 @@ def generate(
...
                     # re-encode it so we get offsets
                     token_offsets = [s for s, e in self.bpe.bpe.encode(text).offsets]
 
                     result = {
                         "text": text,
-                        "tokens": [self.bpe.decode([t]) for t in tokens],
+                        "tokens": [self.bpe.bpe.decode([t]) for t in tokens],
```
ArmenAg, line 329: same here.

ArmenAg:
Are all the file renames working with fb_sweep in metaseq_internal?

lilisierrayu:
> Are all the file renames working with fb_sweep in metaseq_internal?

The only impacted fb_sweep scripts are launch_api_helper.py. Training should be fine. 
Metaseq main already have the rename, so if we have plan to update metaseq_internal, then we should keep the change, otherwise, discard it.

PR #494, opened 2022-11-06
Launch separate sbatch job to copy checkpoints over from scratch to NFS

suchenzang:
Decouples copying from main job, can remove gloo process group if this works.

PR #497, opened 2022-11-08
skip PathManager stuff for NFS save 

zdevito:
It seems to cause problems:

```
line 254, in <lambda>
    self._notify_manager(lambda: self._io.write(b))
OSError: [Errno 14] Bad address
```

But it isn't even needed for saving while also doing training. We can just use concurrent.futures.ThreadPoolExecutor to do the copies from another thread.

Note: how do I verify this doesn't upset the training process timings?


zdevito:
Note I just threw in a revert of #494 to get this to merge. If there are some changes in there we should keep, I can add them back.

suchenzang:
Tested separately with a smol 4 node job, confirmed this works!  Testing with much larger jobs now.

suchenzang:
Works on a 66b run!

suchenzang:
Hm on main right now:
```
2022-11-08 10:34:17 | INFO | metaseq.checkpoint_utils | Saved checkpoint /scratch/slurm_tmpdir/163814/66b_v02_main.bm_none.tokBPE_v3.0_V65536_shard00.bf16.zero2.gelu.transformer_lm_megatron.nlay52.emb10240.lrnpos.0emb_scale.tps2048.adam.b2_0.95.cl1.0.lr0.0001.wu178.dr0.0.atdr0.0.0emb_dr.wd0.1.ms8.mu381469.s1.ngpu1024/checkpoint_250-model_part-0-shard0.pt (epoch 1 @ 250 updates, score 8.201) (writing took 1.222882273956202 seconds)
Submitted batch job 163823
2022-11-08 10:34:26 | INFO | train_inner | {"epoch": 1, "actv_norm": "3520.54", "pos_norm": "0.302", "tok_norm": "0.436", "emb_norm": "0.001", "docsperex": "2.8", "loss": "8.097", "ppl": "273.75", "wps": "1931.4", "ups": "0", "wpb": "2.09715e+06", "bsz": "1024", "num_updates": "251", "lr": "9.99904e-05", "gnorm": "1.31", "clip": "100", "train_wall": "9", "cuda_gb_allocated": "23.2", "cuda_gb_reserved": "53.3", "cuda_gb_free": "56.2", "wall": "3219"}
```

Our NFS write speed is ~25GB/s which seems to match up with needing ~30 seconds to write ~742GB.  This seems to suggest we are still blocking on copying to NFS, though maybe our logging isn't showing the full picture here?

zdevito:
New update with the latest try for this logic:

1. Use `subprocess.run(["cp", ...)` instead of `shutil` to make sure the copy runs in the background.
1. We write `checkpoint_50/_checkpoint_50-shard0.pt` to NFS, then rename removing the _ when done.
1. We check on load that the checkpoint has the appropriate number of non-_ files before considering it a valid checkpoint. This way we do not have to barrier in training process.

suchenzang:
Nit on lint - and good to go!

suchenzang (approved):
Nit on lint, and good to go!

metaseq/checkpoint_utils.py
```diff
@@ -113,14 +111,6 @@ def _copy_if_not_async(src, dest):
             f"Saved checkpoint {checkpoints[0]} (epoch {epoch} @ {num_update} updates) "
```
suchenzang, line 111: Oh something got messed up here in the reverts somehow - had to update this to `update` instead of `num_update`.

PR #500, opened 2022-11-08
Allow to skip batches during training

Xirider:
Allows skipping ranges of batches during training.

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -444,7 +444,7 @@ def post_checkpoint_callback(cfg, filename):
                 filename,
                 cfg.checkpoint.cloud_upload_path,
             ]
-            res = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            res = _run_azcopy(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             if res.returncode != 0:
                 print("Error: {}, azcopy failed".format(res.returncode))
                 print("Azcopy stdout = {}".format(res.stdout))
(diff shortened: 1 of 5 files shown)


metaseq/cli/train.py
```diff
@@ -246,7 +246,8 @@ def train(
                 not cfg.optimization.train_with_epoch_remainder_batch
             ),
         )
-
+    if len(cfg.dataset.skip_batches) > 0:
+        itr = iterators.SkipBatchesIterator(itr, trainer, cfg.dataset.skip_batches)
```
stephenroller, line 250: Calling out that we need to protect this logic from requeues (but only when we get to the next checkpoint) or otherwise have absolute numbers or something

Xirider: What do you think is the issue with requeues?

Reason for using num_updates instead of an absolute number:
We need a way to identify in our logs or tensorboard which batches caused loss spikes. So I'm using the num_updates number, which we get from our logs, to define a batch. I looked at an good way to define absolute numbers that had issues as well.

But I think this should work with restarts from checkpoints and requeues.
Example:
normal order: 1 2 3 4 5 6 7 8 9 10 11 12
skip_batches = "4-6"

order with skipping:
1 2 3 skip(4 5 6) 4 5 6 7 8 9
Now if we restart from checkpoint 3, we immediately skip 4-6 again correctly
if we restart from checkpoint 4 or 5, the data loader will contain the forwarded state with the correct shard and iteration, as it was saved together with the checkpoint. We also won't skip here again, as in our checkpoint 4, the 4th step is already finished, so the 5th is starting now.

stephenroller: Got it, thanks for the clarification

ngoyal2707:
lets not merge this, I tried this locally and I think current way of creating skip iterator always return has_next() False, which makes the code think every update is end_of_epoch. I am using this code but trying other versions

Xirider:
> lets not merge this, I tried this locally and I think current way of creating skip iterator always return has_next() False, which makes the code think every update is end_of_epoch. I am using this code but trying other versions

Mhm, that's weird. For me that didn't happen. But your solution is safer, as it avoids all the iterator wrappers!

PR #501, opened 2022-11-08
Fix verify shards - to fix resuming from checkpoint

sriniiyer:
Currently, num_gpus is set to the local setting of 8, and not the actual number of global gpus. This causes verify-shards to fail and so, resuming from checkpoint fails.

This fixes that

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -179,18 +179,19 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
-def verify_shards(cfg, dir=None, checkpoint_name=None):
+def verify_shards(cfg, dir=None, checkpoint_name=None, distributed_world_size=None):
     # verifies that all the shards of the checkpoint are present
     checkpoint_name = checkpoint_name.replace(".pt", "")
-    num_gpus = DistributedTrainingConfig.distributed_world_size
+    num_gpus = DistributedTrainingConfig.distributed_world_size if not distributed_world_size else distributed_world_size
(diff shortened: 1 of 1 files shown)


metaseq/checkpoint_utils.py
```diff
@@ -179,18 +179,19 @@ def _delete_old_checkpoint_files(
                 os.remove(old_chk)
 
 
-def verify_shards(cfg, dir=None, checkpoint_name=None):
+def verify_shards(cfg, dir=None, checkpoint_name=None, distributed_world_size=None):
     # verifies that all the shards of the checkpoint are present
     checkpoint_name = checkpoint_name.replace(".pt", "")
-    num_gpus = DistributedTrainingConfig.distributed_world_size
+    num_gpus = DistributedTrainingConfig.distributed_world_size if not distributed_world_size else distributed_world_size
```
todpole3, line 185: Why does DistributedTrainingConfig.distributed_world_size differ from distributed_world_size?

Should we set DistributedTrainingConfig.distributed_world_size somewhere instead of creating this extra parameter?

sriniiyer: I'm not sure, but DistributedTrainingConfig seems to be a global pre-set config. I didn't want to mess up anything else by changing it, so I made sure to only affect verify_shards() - which only checks if all the shards are present, which is non-critical and will not introduce any errors. The original metaseq has diverged from this and does not use verify_shards() anymore, so eventually, we can include their commits, but that requires more careful rebasing.

todpole3 (approved):
LGTM, ty!

PR #499, opened 2022-11-08
Save before validating, remove best* checkpoint logic

suchenzang:
Tested:

```
2022-11-09 12:41:47 | INFO | metaseq.checkpoint_utils | Preparing to save checkpoint for epoch 1 @ 10 updates
2022-11-09 12:41:47 | INFO | metaseq.trainer | Saving checkpoint to /scratch/slurm_tmpdir/164255/<redacted>/checkpoint_10-model_part-0-shard0.pt
2022-11-09 12:41:48 | INFO | metaseq.trainer | Finished saving checkpoint to /scratch/slurm_tmpdir/<redacted>/checkpoint_10-model_part-0-shard0.pt
2022-11-09 12:41:48 | INFO | metaseq.checkpoint_utils | Saved checkpoint /scratch/slurm_tmpdir/164255/<redacted>/checkpoint_10-model_part-0-shard0.pt (epoch 1 @ 10 updates) (writing took 1.2194619262591004 seconds)
Submitted batch job 164257
2022-11-09 12:41:48 | INFO | metaseq.cli.train | begin validation on "valid/thread_<redacted>_downsample" subset on rank 0
2022-11-09 12:41:48 | INFO | metaseq.tasks.streaming_language_modeling | setting shuffle buffer size to 10240
```

PR #502, opened 2022-11-09
Split to multiple sbatch copy jobs

suchenzang:
Testing off of https://github.com/facebookresearch/metaseq/pull/499, will rebase onto main after that one is merged.

If only we had pdsh...


metaseq/checkpoint_utils.py
```diff
@@ -163,23 +163,44 @@ def _launch_sbatch_for_checkpoint_copy(
...
+            host_list = (
+                subprocess.check_output(
+                    f"scontrol show hostlist {node_list}", shell=True
                 )
+                .decode()
+                .strip()
             )
-        subprocess.call([f"sbatch {sbatch_run_file}"], shell=True)
+            sbatch_run_file = os.path.join(
+                copy_script_dir, f"_cp_sbatch_script_{num_update}_{host_list}.sh"
```
Xirider, line 189: Does this work? Maybe {idx} instead of {host_list} in the script name to be safe?

metaseq/checkpoint_utils.py
```diff
@@ -163,23 +163,44 @@ def _launch_sbatch_for_checkpoint_copy(
...
-                    oss_dir=oss_dir,
-                    slurm_nodes=slurm_nodes,
-                    local_dir=cfg.save_dir,
-                    num_files=num_files_per_host,
-                    nfs_dir=nfs_upload_path[4:],
-                    cp_script_dir=cp_script_dir,
-                    num_update=num_update,
+        num_hosts_per_chunk = 8
+        for idx in range(
+            0, len(slurm_node_list), num_hosts_per_chunk
```
Xirider, line 178: Can we launch so many jobs if we use a lot of nodes for the training? Could there be an issue if not all are getting resources in time?

Xirider (approved):

metaseq/checkpoint_utils.py
suchenzang: Yup! Just have quotes around filenames at the end of this.

metaseq/checkpoint_utils.py
suchenzang: These are smol - asking for 4 cpu cores each (probably could do with just 1...) and no GPUs. We schedule 16 of them for a 128 host job, and they all end up scheduled on the same node :)

PR #510, opened 2022-11-11
[distributed] Custom rendezvous

stephenroller:
**Patch Description**
The PyTorch TCP rendezvous seems to have trouble if there are minor network blips, causing troubles in getting past initialization at sufficient scale. This patch introduces a custom rendezvous method which copy-pastes much of the pytorch code, and adds a simple retry loop around it.

**Testing steps**
Ongoing


suchenzang (approved):

ngoyal2707 (approved):
Thanks soooooo much for getting it out so fast

PR #515, opened 2022-11-13
added way to skip batches based on Peters previous PR

ngoyal2707:
Based on https://github.com/facebookresearch/metaseq/pull/500

that version was triggering model save every update.

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -193,6 +193,17 @@ def main(cfg: DictConfig) -> None:
         logger.info("PathManager finished waiting.")
 
 
+def get_skip_batches(to_skip):
+    skip_batches = {}
+    for skip_range in to_skip.split(","):
+        if "-" in skip_range:
+            start, end = skip_range.split("-")
+            skip_batches[int(start)] = int(end) - int(start) + 1
+        else:
+            skip_batches[int(skip_range)] = 1
+    return skip_batches
+
+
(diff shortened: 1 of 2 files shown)


metaseq/cli/train.py
```diff
@@ -288,7 +299,23 @@ def train(
...
+    if len(cfg.dataset.skip_batches) > 0:
+        skip_batches = get_skip_batches(cfg.dataset.skip_batches)
     for i, samples in enumerate(progress):
+        current_step = trainer.get_num_updates() + 1
+        if (
+            skip_batches is not None
+            and current_step in skip_batches
+            and skip_batches[current_step] > 0
+        ):
+            skip_batches[current_step] -= 1
```
suchenzang, line 312: Hmm not sure this works - skip_batches is keyed by starting batch index, so if there's more than one batch to skip (passed in range), then by this logic I think we will only ever skip one batch in that range...

suchenzang (approved):
just an accounting question

metaseq/cli/train.py
Xirider: We can either add a new logged metric for each step that shows the current shard and shard iteration of a batch, which would give us an absolute numbering, and skip based on that.

Or use relative numbering with the num_updates that we already have. And we can always afterwards convert the skip_batches list into a absolute list for accounting

PR #517, opened 2022-11-14
Adding relu for seq par

suchenzang:
Testing right now

### DIFF
--- a/metaseq/modules/activation_functions.py
+++ b/metaseq/modules/activation_functions.py
@@ -37,8 +37,7 @@ def relu(x):
 
 @torch.jit.script
 def relu_back(g, x):
-    g[x <= 0] = 0
-    return g
+    return g.masked_fill_(x <= 0, 0)
 
 
 @torch.jit.script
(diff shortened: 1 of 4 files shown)


metaseq/modules/activation_functions.py
```diff
@@ -30,6 +30,17 @@ def gelu_back(g, x):
...
 
+@torch.jit.script
+def relu(x):
+    return F.relu(x)
+
+
+@torch.jit.script
+def relu_back(g, x):
+    g[x <= 0] = 0
+    return g
```
stephenroller, line 41: is it okay to be modifying g inline?

stephenroller: I guess it is because we only use grad_fc2_input for this computation and that's it. Still a bit scary

metaseq/modules/activation_functions.py
```diff
@@ -30,6 +30,17 @@ def gelu_back(g, x):
...
 
 
+@torch.jit.script
+def relu(x):
+    return F.relu(x)
+
+
+@torch.jit.script
+def relu_back(g, x):
+    g[x <= 0] = 0
```
ngoyal2707, line 40: possibly this can add stream synchronization? not really sure, lets just switch to masked_fill_ to be safe?

ngoyal2707 (approved):

stephenroller (approved):

PR #518, opened 2022-11-15
Add a script to benchmark generator

tangbinh:
## Summary
Add a new script to collect latency results for OPT models during generation. While this script resembles the existing one [`metaseq/scripts/generation_benchmarks.py`](https://github.com/facebookresearch/metaseq/pull/240), it's a bit more general where besides latency, we also collect memory usage and GPU traces for various configurations of batch size, input length, and output length. We also use the `GeneratorInterface` directly and skip the checkpoint downloading part.

### DIFF
--- a/metaseq/benchmark/generator.py
+++ b/metaseq/benchmark/generator.py
@@ -1,0 +1,136 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import logging
+import random
+import socket
+import sys
+from typing import Tuple
+
+import numpy as np
+import torch
+from torch.profiler.profiler import (
+    ProfilerActivity,
+    schedule,
+    tensorboard_trace_handler,
+)
+from tqdm import tqdm
+
+from metaseq import options
+from metaseq.dataclass.configs import MetaseqConfig
+from metaseq.dataclass.utils import convert_namespace_to_omegaconf
+from metaseq.distributed import utils as distributed_utils
+from metaseq.hub_utils import GeneratorInterface
+
+logging.basicConfig(format="%(asctime)s | %(name)s | %(message)s", level=logging.INFO)
+logger: logging.Logger = logging.getLogger("metaseq.benchmark.generator")
+
+WARMUP_STEPS = 5
+NUM_REPEATS = 5
+generator = None
+
+
+def main(cfg: MetaseqConfig) -> None:
+    global generator
+    torch.manual_seed(random.randint(1, 20000))
+
+    logger.info("Instantiating a generator interface and loading a model")
+    generator = GeneratorInterface(cfg)
+    generator.load_model()
+
+    batch_size = cfg.dataset.batch_size
+    input_length = cfg.task.max_source_positions
+    output_length = cfg.task.max_target_positions
+
+    logger.info("Running warm-up steps before benchmarking")
+    for _ in range(WARMUP_STEPS):
+        inputs = np.random.randint(100, 50000, size=(batch_size, input_length))
+        generator.generate(inputs, max_tokens=[output_length] * batch_size)
+
+    logger.info(
+        f"Benchmarking with batch size {batch_size}, input length "
+        f"{input_length}, and output length {output_length}"
+    )
+    time_elapsed, peak_memory = benchmark(batch_size, input_length, output_length)
+    if torch.distributed.get_rank() == 0:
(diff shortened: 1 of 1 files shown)


tangbinh:
Please see the updated PR for a script to benchmark the generator interface in terms of latency and peak GPU memory usage. We also add a function to collect GPU traces if needed.

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+    global generator
+    torch.manual_seed(random.randint(1, 20000))
+
+    logger.info("Instantiating a generator interface and loading a model")
+    generator = GeneratorInterface(cfg)
+    generator.load_model()
+
+    batch_size = cfg.dataset.batch_size
+    input_length = cfg.task.max_source_positions
+    output_length = cfg.task.max_target_positions
```
ruanslv, line 45: Lol funny names for input and output length. From fairseq I'm guessing

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+
+    logger.info("Running warm-up steps before benchmarking")
+    for _ in range(WARMUP_STEPS):
+        inputs = np.random.randint(100, 50000, size=(batch_size, input_length))
```
ruanslv, line 49: Where did 100-50000 internal come from? Would it work for any model?

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+    batch_size: int, input_length: int, output_length: int
+) -> Tuple[np.ndarray, np.ndarray]:
+    time_elapsed = []
+    for _ in range(NUM_REPEATS):
+        start = torch.cuda.Event(enable_timing=True)
+        end = torch.cuda.Event(enable_timing=True)
+        start.record()
+
+        generator.generate(
+            inputs=np.random.randint(100, 50000, size=(batch_size, input_length)),
```
ruanslv, line 72: Ditto here on the interval. Should we move to consts?

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+
+    with torch.profiler.profile(
+        activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
+        profile_memory=True,
+        record_shapes=True,
+        schedule=tracing_schedule,
+        on_trace_ready=trace_handler,
+        with_stack=True,
+    ) as prof:
+        for _ in tqdm(range(17)):
```
ruanslv, line 105: 17?

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+            generator.generate(
+                inputs=np.random.randint(100, 50000, size=(batch_size, input_length)),
```
ruanslv, line 107: Ditto on interval

ruanslv (approved):
Very clean script, nice! One optional idea (but can leave for future) would be to use random weights if no model path is provided

cc @erichan1

metaseq/benchmark/generator.py
tangbinh: I think these options are closest to input and output lengths. For context, it's nontrivial to introduce new arguments as we have some validation logic related dataclass fields, which I don't think should be modified for this purpose.

metaseq/benchmark/generator.py
tangbinh: Good point. The OPT models employ a vocabulary whose size exceeds 50K, and I thought the inference time might depend on the inputs (hence the uniform sampling). After testing, I don't think that's the case, though, so let me change it back to fixed prompts.

metaseq/benchmark/generator.py
```diff
@@ -0,0 +1,136 @@
...
+
+    parser = options.get_generation_parser()
+    parser.set_defaults(lr_scheduler=None, criterion=None)
+    LAUNCH_ARGS = ["--task language_modeling", "--bpe hf_byte_bpe", "/tmp"]
+    launch_args = sys.argv[1:] + [item for arg in LAUNCH_ARGS for item in arg.split()]
+    args = options.parse_args_and_arch(parser, input_args=launch_args)
+
+    args.bpe_merges = args.merges_filename
+    args.bpe_vocab = args.vocab_filename
+    cfg = convert_namespace_to_omegaconf(args)
```
erichan1, line 135: I added the following here:
```
model_overrides = {
        "memory_efficient_fp16": True,
        "fp16": True,
        "bf16": True,
        "sequence_parallel": False,
        "inference": True,
        "truncate_init": False
    }
cfg.common_eval.model_overrides = str(model_overrides)
```
Not sure if this is necessary. I can add this option as a follow up PR in case we want to load a bf16 checkpoint.

suchenzang (approved):

PR #521, opened 2022-11-17
Automatically detect and skip loss spikes

Xirider:
**Patch Description**
Added a new flag `max_loss_to_skip_batch` that, if set to some maximum acceptable loss will abort the iteration before doing an optimizer step.
The loss value to compare to is the same one used in the logs. It might be or not different to the one in tensorboard.
The logic is similar to our `skip_gradient_update_on_clip_norm` flag which also skips batches, whenever the gradient norm is above the clip value, and also how we handle overflows.

**Testing steps**
Tested this with our small sweep script. I think our disks are full so I couldn't test this with a longer run.
For testing I increased the loss and checked whether we are skipping correctly

### DIFF
--- a/metaseq/trainer.py
+++ b/metaseq/trainer.py
@@ -141,8 +141,6 @@ class Trainer(object):
         self._previous_training_time = 0
         self._cumulative_training_time = None
 
-        self.real_steps = 0
-
     def reinitialize(self):
         """Reinitialize the Trainer, typically after model params change."""
         self._lr_scheduler = None
@@ -546,9 +544,6 @@ class Trainer(object):
 
             if "ewm_loss" in extra_state:
                 self._ewm_loss = extra_state["ewm_loss"]
-                logger.info(
-                    f"loaded ewm loss from checkpoint with value {self._ewm_loss}"
-                )
 
             self.lr_step(epoch)
 
@@ -983,29 +978,18 @@ class Trainer(object):
         )
 
     def skip_spike(self, logging_outputs, ewm_ratio_to_skip_batch):
-        self.real_steps += 1
         if ewm_ratio_to_skip_batch == -1:
             return
         loss_sum = sum(log.get("loss", 0) for log in logging_outputs)
         sample_size = sum(log.get("sample_size", 0) for log in logging_outputs)
         loss_t = float(loss_sum / sample_size / math.log(2))
 
-        if self.real_steps % 10 == 0:
-            loss_t *= 10
-            logger.info("Increased loss by 10x")
-
         if self._ewm_loss is None:
             self._ewm_loss = loss_t
-            logger.info(
-                "ewm loss was None at this point, so it was initilialized to loss_t"
-            )
 
         ewm_t_1 = self._ewm_loss
         ewm_t = ewm(loss_t, ewm_t_1, span=9)
         ewm_ratio = loss_t / ewm_t
-        logger.info(
-            f"Step {self.get_num_updates()}: loss_t: {loss_t:.2f}, ewm_t_1: {ewm_t_1:.2f}, ewm_t: {ewm_t:.2f}, ewm ratio: {ewm_ratio:.2f}"
-        )
 
         if ewm_ratio > ewm_ratio_to_skip_batch:
             raise SpikeError(
--- a/metaseq/trainer.py
+++ b/metaseq/trainer.py
@@ -979,8 +979,6 @@ class Trainer(object):
         )
 
     def skip_spike(self, logging_outputs, ewm_ratio_to_skip_batch, span=9):
-        if ewm_ratio_to_skip_batch == -1:
-            return
         loss_sum = sum(log.get("loss", 0) for log in logging_outputs)
         sample_size = sum(log.get("sample_size", 0) for log in logging_outputs)
         loss_t = float(loss_sum / sample_size / math.log(2))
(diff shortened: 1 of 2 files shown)


Xirider:
Tested this and everything seems to work now!

I realized that testing this PR by randomly increasing loss with random.randint is not a good idea, as each gpu worker skips a different number of batches 🥲

Xirider:
Also checked for changes in loss and other metrics. This is a 125m run:

1. With this PR and a loss spike every 10 steps:
image

2. with main:
image

Everything looks as expected.

metaseq/trainer.py
```diff
@@ -966,6 +970,19 @@ def clip_grad_norm(
             skip_gradient_update_on_clip_norm=skip_gradient_update_on_clip_norm,
         )
 
+    def skip_spike(self, logging_outputs, max_loss_to_skip_batch):
+        if max_loss_to_skip_batch == -1:
+            return
+        loss_sum = sum(log.get("loss", 0) for log in logging_outputs)
+        sample_size = sum(log.get("sample_size", 0) for log in logging_outputs)
+        loss = float(loss_sum / sample_size / math.log(2))
```
Xirider, line 978: This is the same calculation that happens for logging in reduce_metrics: https://github.com/facebookresearch/metaseq/blob/37809ec65632c434ed722a5c8aa37cd194ce9f8b/metaseq/criterions/cross_entropy.py#L103
The loss value used here matches the one that is logged.

metaseq/trainer.py
```diff
@@ -809,6 +809,10 @@ def maybe_no_sync():
             )
             grad_norm = torch.tensor(0.0).cuda()
             self.zero_grad()
+        except SpikeError as e:
+            overflow = True
+            logger.info(str(e))
+            self.zero_grad()
```
Xirider, line 815: This matches how overflow skips and gradient skips are handled

Xirider:
> Just chatted with @stephenroller on this, and my main q here is: would it be possible to not hardcode a threshold for max loss but rather have it be defined as a function of some "moving average" of losses?
> 
> Ideally we would want to exclude "anomalies" which can be some function of some (exponential) moving average of previous losses.

Yes, I thought about this as well, but then we could have problems with spikes directly after a requeue if our ema is not filled at that point. Also it is more susceptible to weird training fluctuations.

A fixed value has the disadvantages that during the initial training restarts we will need to lower the value manually sometimes. But it seems for our current run, we won't need to do that again.

But I can modify the PR to skip at a multiple of EMA, if you think we want to be more automated :)

Xirider:
This PR now uses a threshold value for the loss to loss ewm ratio to detect loss spikes, instead of a fixed loss value.

### Testing:
add this to sweep_baseline.py
```
hyperparam(
     "--ewm-ratio-to-skip-batch",
     1.25,
),
```

Result logs:
with a loss spike every 10 steps
[P564295826](https://www.internalfb.com/intern/paste/P564295826/)

with a loss spike every 10 steps and restarted from a previous checkpoint
[P564298095](https://www.internalfb.com/phabricator/paste/view/P564298095?lines=395)

metaseq/trainer.py
```diff
@@ -966,6 +977,31 @@ def clip_grad_norm(
...
+        if ewm_ratio > ewm_ratio_to_skip_batch:
+            raise SpikeError(
+                f"Skip batch as we encountered a loss spike. In "
+                f"num_update: {self.get_num_updates()} the loss is {loss_t:.2f}. "
+                f"The ewm for the loss was only at {ewm_t:.2f} . "
+                f"The loss to ewm loss ratio is {ewm_ratio:.2f}, which is higher than "
+                f"ewm_ratio_to_skip_batch of {ewm_ratio_to_skip_batch} ."
+            )
+
+        self._ewm_loss = ewm_t
```
Xirider, line 1003: Note: We are only updating the ewm if we keep the update. Otherwise the ewm would adapt quickly to bad batches and then stop skipping them.

metaseq/trainer.py
```diff
@@ -1245,3 +1281,12 @@ def _set_module_by_path(module, path, value):
     for name in path[:-1]:
         module = getattr(module, name)
     setattr(module, path[-1], value)
+
+
+def ewm(loss_t, ewm_t_1, span):
+    alpha = 2 / (span + 1)
+    return (1 - alpha) * ewm_t_1 + alpha * loss_t
```
stephenroller, line 1288: let's just inline this function please

metaseq/trainer.py
stephenroller: Let's leave a comment explaining this please

metaseq/trainer.py
```diff
@@ -966,6 +977,31 @@ def clip_grad_norm(
...
+            return
+        loss_sum = sum(log.get("loss", 0) for log in logging_outputs)
+        sample_size = sum(log.get("sample_size", 0) for log in logging_outputs)
+        loss_t = float(loss_sum / sample_size / math.log(2))
+
+        if self._ewm_loss is None:
+            self._ewm_loss = loss_t
+
+        ewm_t_1 = self._ewm_loss
+        ewm_t = ewm(loss_t, ewm_t_1, span=9)
```
stephenroller, line 991: is it worth having ewm_loss logged to our metrics?

metaseq/trainer.py
stephenroller: I hope this works with skipping batches from pre-emption but i guess it should.

metaseq/trainer.py
Xirider: Might not be necessary as we can always calculate it again based on the logs? There is no use for the ewm_loss value outside of skipping. When a batch is skipped we are logging ewm_loss.

stephenroller (approved):
thanks, let's give it a try

PR #527, opened 2022-11-22
[tboard] Don't run out of letters

stephenroller:
**Patch Description**
Our restart logic is too good these days. Make sure tboard doesn't ever run out of suffixes.

**Testing steps**
quick job test created the correct tboard folder.


suchenzang (approved):
too many restarts 😂

PR #530, opened 2022-11-24
Don't persist --reset-dataloader past requeues

stephenroller:
**Patch Description**
If we manually pass --reset-dataloader, we need to ignore it after a requeue.

**Testing steps**
Did not test.


suchenzang (approved):

suchenzang:
Gah - didn't catch the case where we are requeue-ing before any new checkpoints are written (reset-dataloader + restore-file), at which point reset dataloader should be persisted...

PR #531, opened 2022-11-24
Default to always skip last batch

suchenzang:
At the moment, we never toggle `train_with_epoch_remainder_batch` anyway - removing this flag for now to reduce surface area here.


stephenroller (approved):
the paranoia is strong

PR #533, opened 2022-11-24
[Cleanup] Checkpoint saving cruft

suchenzang:
Tested with small run with nfs:// upload path, confirmed two sets of checkpoints saved without issue.


Xirider (approved):
Awesome!

PR #535, opened 2022-11-29
barrierless rendezvous for faster init

igormolybogFB:
**Patch Description**
Added a barrierlessenv init_method for pytorch distributed method init_process_group. This is done for faster initialization of the process group

**Testing steps**
Run with --distributed_init_method barrierlessenv://

### DIFF
--- a/metaseq/distributed/rendezvous.py
+++ b/metaseq/distributed/rendezvous.py
@@ -119,5 +119,45 @@ def meet_at_rendezvous(url, **kwargs):
     return iter([(STORE, rank, world_size)])
 
 
+def _tcp_retry_barrierless_rendezvous_handler(url: str, timeout=default_pg_timeout, **kwargs):
+    def _error(msg):
+        return ValueError("barrierlesstcpr:// rendezvous: " + msg)
+
+    global STORE
+
+    result = urlparse(url)
+    if not result.port:
+        raise _error("port number missing")
+    query_dict = _query_to_dict(result.query)
+    if "rank" not in query_dict:
+        raise _error("rank parameter missing")
+    if "world_size" not in query_dict:
+        raise _error("world size parameter missing")
+
+    rank = int(query_dict["rank"])
+    world_size = int(query_dict["world_size"])
+    assert result.hostname is not None
+
+    for tries in range(1, RETRIES + 1):
+        try:
+            store = _create_c10d_store(
+                result.hostname, result.port, rank, world_size, timeout
+            )
+            logger.warning(
+                f"Successfully connected to primary node on attempt {tries}/{RETRIES}"
+            )
+            STORE = ComplicitStore(store, world_size)
+            yield (STORE, rank, world_size)
+            break
+        except RuntimeError as re:
+            logger.error(
+                f"Failed to connect to primary node on attempt {tries}/{RETRIES}: {re}"
+            )
+            time.sleep(COOLDOWN * (2**tries))
+
+    # If this configuration is invalidated, there is nothing we can do about it
+    raise RuntimeError("Unable to perform re-rendezvous using barrierlesstcpr:// method")
+
 register_rendezvous_handler("barrierlessenv", meet_at_rendezvous)
 register_rendezvous_handler("tcpr", _tcp_retry_rendezvous_handler)
+register_rendezvous_handler("barrierlesstcpr", _tcp_retry_barrierless_rendezvous_handler)
--- a/metaseq/distributed/rendezvous.py
+++ b/metaseq/distributed/rendezvous.py
@@ -120,43 +120,13 @@ def meet_at_rendezvous(url, **kwargs):
 
 
 def _tcp_retry_barrierless_rendezvous_handler(url: str, timeout=default_pg_timeout, **kwargs):
-    def _error(msg):
-        return ValueError("barrierlesstcpr:// rendezvous: " + msg)
-
+    assert url.startswith("barrierlesstcpr://")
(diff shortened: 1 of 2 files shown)


suchenzang (commented):
yesss! one question though: can we combine both this rendezvous with the retry logic in tcpr ?

igormolybogFB:
@suchenzang do we have examples of usage of tcpr in our main?

metaseq/distributed/rendezvous.py
```diff
@@ -70,4 +70,94 @@ def _error(msg):
...
+            yield (STORE, rank, world_size)
+            break
+        except RuntimeError as re:
+            logger.error(
+                f"Failed to connect to primary node on attempt {tries}/{RETRIES}: {re}"
+            )
+            time.sleep(COOLDOWN * (2**tries))
+
+    # If this configuration is invalidated, there is nothing we can do about it
+    raise RuntimeError("Unable to perform re-rendezvous using barrierlesstcpr:// method")
```
lw, line 159: I think this can be simplified substantially to something like this:
```suggestion
    assert url.startswith("barrierlesstcpr://")
    url = url.replace("barrierlesstcpr://", "tcpr://")
    rendezvous_iterator = rendezvous(url)
    store, rank, world_size = next(rendezvous_iterator)
    global STORE
    STORE = ComplicitStore(store, world_size)
    return iter([(STORE, rank, world_size)])
```

lw: And, in fact, it could even further be merged with the `meet_at_rendezvous` function above, to deduplicate the two almost-identical functions.

lw (approved):

suchenzang (approved):

stephenroller:
i don't understand how this works. can you explain?

metaseq/distributed/rendezvous.py
```diff
@@ -70,4 +70,52 @@ def _error(msg):
...
+    def set(self, *args, **kwargs):
+        return self.store.set(*args, **kwargs)
+
+    def get(self, *args, **kwargs):
+        return self.store.get(*args, **kwargs)
+
+    def add(self, key: str, *args, **kwargs):
+        # The marker isn't always a prefix: it's sometimes prefixed with default_pg/.
+        if STORE_BASED_BARRIER_MARKER in key:
+            return self.world_size
```
stephenroller, line 91: I see, specifically if it's the initialization, we just pretend it's done.

I noticed this part of the docs on TCPStore:

https://pytorch.org/docs/stable/distributed.html#torch.distributed.TCPStore

Namely, on construction it has a `wait_for_worker=True` default parameter, which we are not currently setting. Maybe we can get the same effect by just turning this off?

stephenroller: I see, there's actually two barriers.

One in C++, for TCP store:

https://github.com/pytorch/pytorch/blob/c18da597e0bb1c1aecc97c77a73fed1849057fa4/torch/csrc/distributed/c10d/TCPStore.cpp#L1012-L1050

And one in the process group initialization:
https://github.com/pytorch/pytorch/blob/master/torch/distributed/distributed_c10d.py#L851-L863

And then a third in our own code:
https://github.com/facebookresearch/metaseq/blob/2e882aafef307b200728759dc7e0544622bf4b8c/metaseq/distributed/utils.py#L224

So maybe we can combine both TCPStore(wait_for_worker=False) and this and get even better overlap?

igormolybogFB: @stephenroller what do you want to combine with TCPStore, once again?

As far as I understand, you want to just add `wait_for_workers=False` in _create_c10d_store (https://github.com/facebookresearch/metaseq/blob/2e882aafef307b200728759dc7e0544622bf4b8c/metaseq/distributed/rendezvous.py#L25) and continue using tcpr?  

It looks to me like that should work. @lw what do you think?

lw: I think @stephenroller meant to do both this PR and pass the `wait_for_worker=False` flag to the TCPStore constructor, since they control two different barriers (both of which could be a problem).

I believe the `wait_for_workers=False` flag will have limited effect, because it only controls one barrier which happens in the constructor of the store, and the store is only constructed once! (The sub-PGs re-use the same store as the root PG, they don't create a new one). I do not know why that barrier was needed and what safety implications come with removing it, I'd suggest looking into that.

I'd personally recommend removing each of these barriers in a separate commit, to test their performance and correctness impact separately.

stephenroller: +1 to separating the proposed wait_for_worker=False into an entirely separate commit.

metaseq/distributed/rendezvous.py
```diff
@@ -70,4 +70,59 @@ def _error(msg):
...
+        return self.store.set_timeout(*args, **kwargs)
+
+    def wait(self, *args, **kwargs):
+        return self.store.wait(*args, **kwargs)
+
+
+STORE = None
+
+
+def _tcp_retry_barrierless_rendezvous_handler(
```
suchenzang, line 113: Testing this now - is there retry logic here? I don't see it...

suchenzang: Oh nvm, I see the tcpr string replacement below now, lol

lw: This function calls into the `rendezvous` function (on line 118) using a URL that starts with `tcpr://`, which should thus in turn be rerouted to the retry logic. Does that make sense?

PR #537, opened 2022-11-30
Support overriding sequence parallelism in the API

punitkoura:
**Patch Description**
1. Support overriding sequence parallelism in the API
2. Support loading FSDP sharded models through the API

**Testing steps**

1. Take a sequence parallel checkpoint, and create a constants module file (custom_constants_module.py)

```
# Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

import os

MAX_SEQ_LEN = 2048
BATCH_SIZE = 2048  # silly high bc we dynamically batch by MAX_BATCH_TOKENS
MAX_BATCH_TOKENS = 3072
DEFAULT_PORT = 6010
MODEL_PARALLEL = <REPLACE_MODEL_PARALLEL_SIZE_HERE>
TOTAL_WORLD_SIZE = <REPLACE_WORLD_SIZE_HERE>
MAX_BEAM = 16

CHECKPOINT_FOLDER = <INSERT_MODEL_CHECKPOINT_FOLDER_HERE>

# tokenizer files
HF_TOKENIZER = <INSERT_TOKENIZER_FILE_HERE>
MODEL_FILE = os.path.join(CHECKPOINT_FOLDER, "reshard.pt")


LAUNCH_ARGS = [
    f"--model-parallel-size {MODEL_PARALLEL}",
    f"--distributed-world-size {TOTAL_WORLD_SIZE}",
    "--ddp-backend fully_sharded",
    "--task language_modeling",
    "--bpe hf_byte_bpe",
    f"--hf-tokenizer {HF_TOKENIZER}",
    f"--path {MODEL_FILE}",
    "--beam 1 --nbest 1",
    "--distributed-port 13000",
    "--checkpoint-shard-count 1",
    "--use-sharded-state",
    f"--batch-size {BATCH_SIZE}",
    f"--buffer-size {BATCH_SIZE * MAX_SEQ_LEN}",
    f"--max-tokens {BATCH_SIZE * MAX_SEQ_LEN}",
    "/tmp",  # required "data" argument.
]

# Optional arg overrides which influence model loading during inference
INFERENCE_ARG_OVERRIDES = {"sequence_parallel": False}

```

2. Export the constants module for it to be visible to interactive_hosted.py

```
export PYTHONPATH=$PYTHONPATH:/path/to/custom_constants_module METASEQ_SERVICE_CONSTANTS_MODULE=custom_constants_module
```

3. Run interactive_hosted.py
```
srun --exclusive -N <NUMBER_OF_NODES> --gpus-per-node <NUMBER_OF_GPUS_PER_NODE> --tasks 1 -c 96 --partition <PUT_PARTITION_NAME_HERE> --time "1-00:00:00" --qos high  --pty python metaseq/cli/interactive_hosted.py
```
NUMBER_OF_NODES*NUMBER_OF_GPUS_PER_NODE should be equal to MODEL_PARALLEL_SIZE

4. Prompt the model
```
curl -k http://<ip>:<port>/completions -H "Authorization: Bearer Punit" -H "Content-Type: application/json" \
-d '{
"prompt": [16853, 16947, 19678, 16709, 16647, 19493, 17495, 16688, 16397],
"temperature": 0.0,
"max_tokens": 0, "min_tokens": 0,
"top_p": 1.0, "n": 1, "best_of": 1,
"echo": true, "logprobs": 1, "seed": 1
}'
```

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -18,7 +18,11 @@ from omegaconf import OmegaConf
 from metaseq.dataclass.configs import CheckpointConfig
 from metaseq.dataclass.utils import overwrite_args_by_name
 from metaseq.distributed import utils as distributed_utils
-from metaseq.file_io import PathManager, torch_load_cpu
+from metaseq.file_io import (
+    PathManager,
+    torch_load_cpu,
+    load_and_pop_last_optimizer_state,
+)
 from metaseq.launcher.opt_job_constants import ComputeEnvs
 
 logger = logging.getLogger(__name__)
@@ -316,7 +320,8 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
             "--model-parallel. If you are working on a new script, it may also mean "
             "you failed to fsdp_wrap or you have an unnecessary fsdp_wrap."
         )
-    sd = torch_load_cpu(checkpoint_files[0])
+    # We don't need the optimizer state here, since we're just loading the model for the config.
+    sd = load_and_pop_last_optimizer_state(checkpoint_files[0])
     return sd["cfg"]["distributed_training"]["use_sharded_state"]
 
 
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -18,11 +18,7 @@ from omegaconf import OmegaConf
 from metaseq.dataclass.configs import CheckpointConfig
 from metaseq.dataclass.utils import overwrite_args_by_name
 from metaseq.distributed import utils as distributed_utils
-from metaseq.file_io import (
-    PathManager,
-    torch_load_cpu,
-    load_and_pop_last_optimizer_state,
-)
+from metaseq.file_io import PathManager, torch_load_cpu
 from metaseq.launcher.opt_job_constants import ComputeEnvs
 
 logger = logging.getLogger(__name__)
@@ -320,8 +316,7 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
             "--model-parallel. If you are working on a new script, it may also mean "
             "you failed to fsdp_wrap or you have an unnecessary fsdp_wrap."
         )
-    # We don't need the optimizer state here, since we're just loading the model for the config.
-    sd = load_and_pop_last_optimizer_state(checkpoint_files[0])
+    sd = torch_load_cpu(checkpoint_files[0])
     return sd["cfg"]["distributed_training"]["use_sharded_state"]
 
 
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -7,6 +7,7 @@ import argparse
 import ast
 import logging
 import os
+import re
 import time
 from argparse import Namespace
 from typing import List, Optional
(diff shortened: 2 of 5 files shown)


metaseq/hub_utils.py
```diff
@@ -106,6 +108,24 @@ def decode_fn(self, x: List[int]) -> str:
         return self.bpe.bpe.decode(x)
 
     def load_model(self):
+        if self.cfg.common.model_parallel_size == 1:
```
punitkoura, line 111: Support loading FSDP sharded models directly through the API.

metaseq/checkpoint_utils.py
```diff
@@ -316,7 +320,8 @@ def _is_checkpoint_sharded(checkpoint_files) -> bool:
             "--model-parallel. If you are working on a new script, it may also mean "
             "you failed to fsdp_wrap or you have an unnecessary fsdp_wrap."
         )
-    sd = torch_load_cpu(checkpoint_files[0])
+    # We don't need the optimizer state here, since we're just loading the model for the config.
+    sd = load_and_pop_last_optimizer_state(checkpoint_files[0])
```
moyapchen, line 324: Dumb q - is there a particular reason for popping optimizer state in a "is checkpoint sharded" function rather than something more inference-specific? 

(I get that we don't need the optimizer state while loading the model for generation, but `checkpoint_utils.py` is also used to load models for training, yeah?)

metaseq/service/constants.py
```diff
@@ -56,3 +56,6 @@
     f"--max-tokens {BATCH_SIZE * MAX_SEQ_LEN}",
     "/tmp",  # required "data" argument.
 ]
+
+# Optional arg overrides which influence model loading during inference
+ARG_OVERRIDES = {}
```
moyapchen, line 61: Maybe name this to be a slightly more specific "INFERENCE_ARG_OVERRIDES" or something, rather than "ARG_OVERRIDES", so it'll be slightly less confusing reading through code what this is?

metaseq/hub_utils.py
```diff
@@ -120,6 +140,7 @@ def load_model(self):
         task = tasks.setup_task(self.cfg.task)
 
         def _build_model(cfg, task):
+            cfg.model.distribute_checkpointed_activations = False
```
moyapchen, line 143: Dumb q - is there a reason for this line to go here rather than `make_generation_fast_()`?

(I'm assuming it's a load ordering thing... asking cause wondering if there's some nice way to bunch all of these generation-specific override things together — for example, would it make sense to have this live as a default in ARG_OVERRIDES below, or something along those lines?

moyapchen (commented):
Thanks for the PR - couple structure questions.

If you also wouldn't mind copy/pasting the command you run to validate this in the test plan, that'd be great

metaseq/checkpoint_utils.py
punitkoura: Yeah, that's a good point... We use the is_sharded information to identify how the model is to be loaded, and we need to see the config for doing that. And the only way to load the config right now is to load the whole .pt file... (Which we were doing earlier using torch_load_cpu.

The change I made just drops the optimizer state after loading to reduce some memory overhead.

metaseq/service/constants.py
punitkoura: Good point! Let me make this change

metaseq/hub_utils.py
punitkoura: Good point! This can be moved to INFERENCE_ARG_OVERRIDES. Turns out, I did not have to do this if I turn off sequence_parallel. So removing this for now.

moyapchen (approved):
tyty

PR #539, opened 2022-11-30
Debug logging for end_of_epoch

zdevito:
In larger runs some dataloaders appear to end early. Adding some debug logging when they reach an end_of_epoch state to try to spot the difference in these instances vs others.

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -275,7 +275,7 @@ def train(
 
         end_of_epoch = not itr.has_next()
         if end_of_epoch:
-            end_of_epoch = distributed_utils.get_global_rank()
+            grank = distributed_utils.get_global_rank()
             dataset = epoch_itr.dataset
             while not hasattr(dataset, 'len_cache'):
                 dataset = dataset.dataset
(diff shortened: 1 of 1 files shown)


metaseq/cli/train.py
```diff
@@ -274,6 +274,16 @@ def train(
                 metrics.reset_meters("train_inner")
 
         end_of_epoch = not itr.has_next()
+        if end_of_epoch:
+            end_of_epoch = distributed_utils.get_global_rank()
```
stephenroller, line 278: I think you mean
```suggestion
            grank = distributed_utils.get_global_rank()
```

metaseq/cli/train.py
```diff
@@ -274,6 +274,16 @@ def train(
                 metrics.reset_meters("train_inner")
 
         end_of_epoch = not itr.has_next()
+        if end_of_epoch:
```
stephenroller, line 277: can we leave a comment pointing to this PR or the task and saying we want to delete this after we resolve the issue?

zdevito: Should we merge it, or do you just want to cherry-pick it into a job run? I am fine either way

suchenzang (approved):
Nit on lint - @stephenroller are you pulling this into our job branch?

stephenroller:
Pulled into bigbig manually.

PR #542, opened 2022-12-02
Automatically detect and skip loss spikes

Xirider:
Allows to set a threshold for loss to ewa loss ratio, on which to skip a batch. This is useful for skipping loss spikes.
Here are more details: https://github.com/facebookresearch/metaseq/pull/521


suchenzang (approved):
I think if you rebase onto main, the gpu tests should hopefully all pass...

Issue #516, opened 2022-11-13
Gpu test broken on main since #497


Xirider:
A GPU test is broken on main since https://github.com/facebookresearch/metaseq/pull/497

CircleCI output:
```
self = <test_checkpoint_saving.TestCheckpointSavingAndUploading testMethod=test_checkpoint_saving_and_uploading>

    def test_checkpoint_saving_and_uploading(self):
        max_update_first_run = 20
        multiprocessing.set_start_method("spawn", force=True)
        with torch.multiprocessing.Manager() as manager:
            events = manager.list()
            p = multiprocessing.Process(
                target=run_training,
                args=(
                    events,
                    max_update_first_run,
                ),
            )
            p.start()
            p.join()
            events_first_run = list(events)
    
        # check that training ran correctly
        training_log_events = [
            json.loads(event["message"])
            for event in events_first_run
            if event["type"] == "log" and event["message"].startswith('{"epoch"')
        ]
        self.assertEqual(len(training_log_events), max_update_first_run)
        self.assertEqual(
            int(training_log_events[-1]["num_updates"]), max_update_first_run
        )
        self.assertAlmostEqual(float(training_log_events[-1]["loss"]), 14.574, 2)
    
        # check that the correct checkpoints were created and uploaded
        upload_events = [
            event for event in events_first_run if event["type"] == "upload"
        ]
>       self.assertEqual(len(upload_events), 8)
E       AssertionError: 7 != 8

gpu_tests/test_checkpoint_saving.py:60: AssertionError
```

suchenzang:
This test needs to be reworked now to test for async checkpoint saving (by checking presence of checkpoint in test-checkpoint dir instead of proxy-ing the correct number of uploads via log lines).

PR #543, opened 2022-12-03
Fix test checkpoint saving

zycalice:
**Patch Description**
related to issue https://github.com/facebookresearch/metaseq/issues/516
1. fixed distributed_main_mock by patching save_checkpoint in trainer
2. replaced partial with partial method for patching on class method (the log_to_events mock now print out message instead of info due to the presence of self
3. note that now loss matching is changed to matching to the 1st digit

**Testing steps**
Ran python metaseq/gpu_tests/test_checkpoint_saving.py
output:
.
Ran 1 test in 52.583s

OK


metaseq/cli/train.py
```diff
@@ -277,12 +277,22 @@ def train(
...
+                " ".join(
+                    [
+                        f"End of Epoch on rank {grank}:",
+                        f"sequences_consumed={itr.sequences_consumed}",
+                        f"n={itr.n}",
+                        f"len_cache_hash={cache_hash}",
+                        f"len_cache_has_zeros={contains_zero}",
+                    ]
+                )
+            )
```
Xirider, line 295:
```suggestion
            logger.warning(
                f"End of Epoch on rank {grank}: "
                f"sequences_consumed={itr.sequences_consumed} "
                f"n={itr.n} "
                f"len_cache_hash={cache_hash} "
                f"len_cache_has_zeros={contains_zero}"
            )
```
This is how we usually format multiline strings.

Xirider (commented):
Awesome that you managed to fix this!

Xirider (approved):

PR #547, opened 2022-12-07
Scale LR instead

suchenzang:


metaseq/trainer.py
```diff
@@ -777,8 +777,10 @@ def maybe_no_sync():
             # downscale grads by ewm_loss_ratio ** 4
             if ewm_loss_ratio > 1.0:
                 grad_mult_factor = 1.0 / (ewm_loss_ratio ** 4)
-                self.optimizer.multiply_grads(grad_mult_factor)
-                logger.info(f"Scaling grads by {grad_mult_factor:.2f}")
+                curr_lr = self.optimizer.get_lr()
+                new_lr = curr_lr * grad_mult_factor
+                self.optimizer.set_lr(new_lr)
```
stephenroller, line 782: are we absolutely certain this only sets it for one step?

stephenroller (approved):

metaseq/trainer.py
suchenzang: ya we recompute based on num update each time https://github.com/facebookresearch/metaseq/blob/bf71e53ca5459d0fd07b025f997b24e3cddb5331/metaseq/optim/lr_scheduler/polynomial_decay_schedule.py#L82

PR #549, opened 2022-12-11
add log_training_trajectory option

todpole3:
if set, evaluate and save checkpoints more frequently in early stage of training

### DIFF
--- a/metaseq_cli/train.py
+++ b/metaseq_cli/train.py
@@ -406,6 +406,10 @@ def validate_and_save(
             and num_updates >= cfg.dataset.validate_after_updates
             and was_successful_step
         )
+        or (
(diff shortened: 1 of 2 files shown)


metaseq_cli/train.py
```diff
@@ -418,6 +418,10 @@ def validate_and_save(
             and num_updates >= cfg.dataset.validate_after_updates
             and was_successful_step
         )
+        or (
+            cfg.dataset.log_training_trajectory
```
sriniiyer, line 422: why do we need to save? Just validate?

todpole3: I want to plot the trajectory for generation tasks as well (ROUGE); eval will be fast

PR #550, opened 2022-12-12
Automatically evaluate checkpoints after copying to NFS

Xirider:
This PR adds the ability to automatically evaluate checkpoints after copying to NFS.

- After the asynchronous copy of a checkpoint to NFS, we check periodically whether all checkpoint parts have finished copying.
- Once all parts are done, we run the eval script defined in `nfs_eval_script_path` for this checkpoint

Add these lines to the internal sweep script to define the eval script path:
https://www.internalfb.com/phabricator/paste/view/P576097431

Testing:
I tested this with a dummy eval script, as the Punit's real eval script should only be used for large runs. I would recommend to check first whether the eval script alone runs correctly on our frozen branch.

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -408,7 +408,9 @@ def validate_and_save(
             trainer,
             epoch_itr,
             training_finished=should_stop,
-            async_callback_fn=functools.partial(post_checkpoint_callback, cfg)
+            async_callback_fn=functools.partial(
+                post_checkpoint_callback, cfg, num_updates, should_stop
+            )
             if cfg.checkpoint.cloud_upload_path
             else None,
         )
@@ -428,7 +430,7 @@ def _checkpoint_add_directory(basename):
     return m[1], f"checkpoint{m[3]}"
 
 
-def post_checkpoint_callback(cfg, filename):
+def post_checkpoint_callback(cfg, num_updates, should_stop, filename):
     if cfg.checkpoint.cloud_upload_path is not None:
         if "blob.core.windows.net" in cfg.checkpoint.cloud_upload_path:
             azcopy_logs = filename + "_azcopy_logs"
@@ -496,55 +498,14 @@ def post_checkpoint_callback(cfg, filename):
             )
             os.remove(filename)
 
-            # Run evals after all checkpoint parts have been uploaded
-            checkpoint_id = checkpoint_dir.split("_")[-1]
-            if (
-                cfg.checkpoint.cloud_eval_script_path is not None
-                and distributed_utils.get_global_rank() == 0
-                and checkpoint_id.isdigit()
-                and cfg.checkpoint.cloud_eval_frequency > 0
-                and (int(checkpoint_id) % cfg.checkpoint.cloud_eval_frequency == 0)
-            ):
-                for retry in range(cfg.checkpoint.cloud_eval_num_retries):
-                    time.sleep(cfg.checkpoint.cloud_eval_retry_wait_minutes * 60)
-
-                    current_checkpoint_path = os.path.join(
-                        destination_checkpoints_dir, checkpoint_dir
-                    )
-                    num_files = os.listdir(current_checkpoint_path)
-                    finished_checkpoint_parts = len(
-                        [f for f in num_files if not f.startswith("_")]
-                    )
-                    if (
-                        finished_checkpoint_parts
-                        == cfg.distributed_training.distributed_world_size
-                    ):
-                        logger.info(
-                            f"All checkpoint parts for {checkpoint_dir} are in NFS, will now start to run evals"
-                        )
-                        script_dir = os.path.join(
-                            os.environ.get("METASEQ_SAVE_DIR"),
-                            cfg.checkpoint.cloud_eval_script_path,
-                        )
-                        res = subprocess.run(
-                            [
-                                "bash",
-                                script_dir,
-                                os.path.join(current_checkpoint_path, "checkpoint.pt"),
-                            ],
-                            stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE,
-                        )
-                        if res.returncode == 0:
-                            logger.info(f"Sucessfully evaluated {checkpoint_dir}")
-                        else:
-                            logger.info(f"Error during evaluation: {res.returncode}")
-                            logger.info(f"Eval script stdout = {res.stdout}")
-                            logger.info(f"Eval script stderr = {res.stderr}")
-                        return
-                logger.info(
-                    f"Did not evaluate {checkpoint_dir}, as not all checkpoints were copied to NFS within waiting time"
-                )
+            # Start running evals on uploaded checkpoint
+            nfs_evaluation(
+                cfg,
+                num_updates,
+                should_stop,
+                checkpoint_dir,
+                destination_checkpoints_dir,
+            )
 
         else:
             try:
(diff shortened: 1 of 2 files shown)


metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
                 ),
             )
             os.remove(filename)
+
+            # Run evals after all checkpoint parts have been uploaded
+            checkpoint_id = checkpoint_dir.split("_")[-1]
+            if (
+                cfg.checkpoint.cloud_eval_script_path is not None
+                and distributed_utils.get_global_rank() == 0
```
Xirider, line 503: Noob question: When running on many nodes, does this only run once?

metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
                 ),
             )
             os.remove(filename)
+
+            # Run evals after all checkpoint parts have been uploaded
+            checkpoint_id = checkpoint_dir.split("_")[-1]
```
ruanslv, line 500: Could we move this logic to a separate method and delete the existing `_run_evaluations` if it's not used anymore?

metaseq/cli/train.py
ruanslv: Yes

metaseq/dataclass/configs.py
```diff
@@ -560,6 +560,32 @@ class CheckpointConfig(MetaseqDataclass):
             "argparse_alias": "--cloud-dir",
         },
     )
+    cloud_eval_script_path: Optional[str] = field(
```
ruanslv, line 563: We are calling these flags `cloud` but we are not using them for blob nor S3. Could we either make the logic generic or call them `nfs_` instead?

ruanslv: Also, if we are adding new flags can we remove the `evaluate_` ones added previously? https://github.com/facebookresearch/metaseq/pull/411

metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
...
+            if (
+                cfg.checkpoint.cloud_eval_script_path is not None
+                and distributed_utils.get_global_rank() == 0
+                and checkpoint_id.isdigit()
```
ruanslv, line 504: This logic is brittle, could we rely on `num_updates` instead? Also we probably want evals at the end of the training run? See approach in https://github.com/facebookresearch/metaseq/pull/411/files

metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
...
+                        [f for f in num_files if not f.startswith("_")]
+                    )
+                    if (
+                        finished_checkpoint_parts
+                        == cfg.distributed_training.distributed_world_size
+                    ):
+                        logger.info(
+                            f"All checkpoint parts for {checkpoint_dir} are in NFS, will now start to run evals"
+                        )
+                        script_dir = os.path.join(
```
ruanslv, line 525: I don't feel particularly strong either way, but @stephenroller had suggested that we use a python function instead, see comments in https://github.com/facebookresearch/metaseq/pull/411/files. Might be easier for version control?

ruanslv:
Thanks for taking this on! I've added a bunch of comments related to previous PR: https://github.com/facebookresearch/metaseq/pull/411/files

I feel perhaps reusing `_run_evaluations` method and just updating the wait criteria (`sleep` vs `monitored_barrier`) may be easier, but your call if you want to replace it instead. We should keep just one evals method around.

ruanslv (changes requested):

metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
...
+                            stderr=subprocess.PIPE,
+                        )
+                        if res.returncode == 0:
+                            logger.info(f"Sucessfully evaluated {checkpoint_dir}")
+                        else:
+                            logger.info(f"Error during evaluation: {res.returncode}")
+                            logger.info(f"Eval script stdout = {res.stdout}")
+                            logger.info(f"Eval script stderr = {res.stderr}")
+                        return
+                logger.info(
```
stephenroller, line 545: warning? error?

metaseq/cli/train.py
stephenroller: yes plz

metaseq/cli/train.py
```diff
@@ -495,6 +495,57 @@ def post_checkpoint_callback(cfg, filename):
...
+                        logger.info(
+                            f"All checkpoint parts for {checkpoint_dir} are in NFS, will now start to run evals"
+                        )
+                        script_dir = os.path.join(
+                            os.environ.get("METASEQ_SAVE_DIR"),
+                            cfg.checkpoint.cloud_eval_script_path,
+                        )
+                        res = subprocess.run(
+                            [
+                                "bash",
```
stephenroller, line 531: Don't we submit a job? and do this async?

metaseq/dataclass/configs.py
Xirider: You are right. It is is also coupled to NFS, as we are checking for existence of files, and the eval script won't work on non NFS files. 
- will change it to `nfs_`.
- remove evaluate flag

metaseq/cli/train.py
Xirider: Do you mean that instead of running a bash script, we run a python function? Mhm, but we are currently using Punit's bash script for these evals. So we might need to translate that into python first  and validate it again. Not sure that is worth it?

metaseq/cli/train.py
Xirider: What do mean by this?
As we are inside a new Thread for the copy to NFS, all this is done async from the training process already. The bash script is started and submits some jobs, then waits for checkpoint consolidation, submits more jobs for evals.

metaseq/cli/train.py
ruanslv: Yea if we are already using the bash script everywhere then I agree to keep it this way, at least for now

metaseq/cli/train.py
Xirider: - updated to use num_updates
- kept the logic for determining when to evaluate outside, as unlike your implementation its only for nfs
- I'm not sure we actually want to have an extra flag for evals at the end of training. Susan previously wanted me to remove similar logic

metaseq/cli/train.py
```diff
@@ -510,28 +521,59 @@ def post_checkpoint_callback(cfg, filename):
...
-    if not hasattr(module, "eval_fn"):
-        raise RuntimeError(
-            f"{eval_module} must have a function called eval_fn to utilize for evaluations. "
-            "It expects the following signature:\n"
-            "def eval_fn(cloud_upload_path: str, checkpoint_name: str)"
+    if (
+        cfg.checkpoint.nfs_eval_script_path is not None
+        and distributed_utils.get_global_rank() == 0
+        and cfg.checkpoint.nfs_eval_frequency > 0
+        and num_updates % cfg.checkpoint.nfs_eval_frequency == 0
```
ruanslv, line 531: Did you mean to use `should_stop` here as an additional check? 

I find it weird not having a flag to control whether we run at the end, are you just going to turn it on by default then? It seems clear to me that we would want to evaluate last checkpoint

cc @suchenzang Peter mentioned you argued in favour of not having flags for last_checkpoint?

ruanslv (approved):
Overall looks good, added some comments for small adjustments but approving to unblock

metaseq/cli/train.py
suchenzang: yeah hopefully no flag and default to just checkpoint the last checkpoint (which we should be doing now since we checkpoint every epoch)? is there a case where the last step is not also the end of an epoch/data shard?

metaseq/cli/train.py
```diff
@@ -510,28 +521,59 @@ def post_checkpoint_callback(cfg, filename):
...
+    ):
+        for retry in range(cfg.checkpoint.nfs_eval_num_retries):
```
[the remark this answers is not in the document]
Xirider: Good catch! Yes, will change that.

metaseq/cli/train.py
```diff
@@ -510,28 +521,59 @@ def post_checkpoint_callback(cfg, filename):
...
+            time.sleep(cfg.checkpoint.nfs_eval_retry_wait_minutes * 60)
+
+            current_checkpoint_path = os.path.join(
+                destination_checkpoints_dir, checkpoint_dir
+            )
+            num_files = os.listdir(current_checkpoint_path)
+            finished_checkpoint_parts = len(
+                [f for f in num_files if not f.startswith("_")]
```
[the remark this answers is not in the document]
Xirider: Yeah, I will explain this also in a comment. In the NFS copy code above we are first saving the checkpoint part with a leading "_", and once the copying is finished, we rename it by removing the "_". This way we make sure we don't consider half saved checkpoints.

metaseq/cli/train.py
Xirider: - Added option 1: eval last checkpoint by default, assuming nfs_eval_script_path is set.
- renamed as suggested to training_finished, that makes it clearer!

PR #551, opened 2022-12-12
Updates from latest runs

suchenzang:


metaseq/trainer.py
```diff
@@ -774,9 +777,18 @@ def maybe_no_sync():
             ewm_loss_ratio = self.skip_spike(
                 logging_outputs, self.cfg.optimization.ewm_ratio_to_skip_batch
             )
+            # downscale grads by ewm_loss_ratio ** 4
+            if ewm_loss_ratio > 1.0:
```
Xirider, line 781: This means, that we are almost always downscaling the LR, should we not use a bit less aggressive threshold, so we only do the scaling when we reach an actual spike?

metaseq/trainer.py
```diff
@@ -1176,6 +1189,38 @@ def is_consistent(tensor):
...
+            if (not self.is_fsdp) or self.dont_log_param_and_grad_norm:
+                break
+            if param.grad is None:
+                continue
+            nice_name = name.replace("module._fsdp_wrapped_module._fpw_module.", "")
+            nice_name = nice_name.replace("_fsdp_wrapped_module._fpw_module.", "")
+            nice_name = nice_name.replace("._fsdp_wrapped_module.flat_param_0", "")
+            nice_name = nice_name.replace("decoder.layers.", "layer")
+            # threshold for near zeros
+            threshold = torch.finfo(param.grad.dtype).tiny * 2
```
Xirider, line 1203: This threshold seems really low? Do we not want to check for things close to zero as well?

Xirider (approved):

metaseq/trainer.py
suchenzang: ah this was the threshold during FP16 OPT days... keeping it around for now but sufficiently close to 0 here would be good either way

metaseq/trainer.py
suchenzang: I think our downscaling logic still isn't aggressive enough :( and technically we would probably be rolling back to previous checkpoints since it looks to still be too "delayed" when it gets triggered... leaving for now though to get bigbig up to date with what is currently running, and we can keep iterating :)

PR #555, opened 2022-12-21
Add optional type conversion to FSDP resharding script

tangbinh:
### Summary of Changes
We add an option to convert weights into a new `dtype` while resharding FSDP checkpoints. This helps reduce checkpoint sizes and avoids issues under RAM constraints when we attempt to load checkpoints. For example, model weights might be saved in full precision but one only needs half precision at inference time.

We also rename some options (e.g. `input-glob-pattern` → `input` and `output-shard-name` → `output`) for succinctness.

### Test Plan
* Reshard and convert the OPT-125M checkpoint into various dtypes and make sure we can do inference correctly in the new dtypes:
```
seq 0 1 | parallel --line-buffer 'python metaseq/scripts/reshard_fsdp.py --input "/data/checkpoints/opt-125m/raw/checkpoint_last-model_part-{}-shard*.pt" --output "/data/checkpoints/opt-125m/reshard-no-os/reshard-model_part-{}.pt" --skip-optimizer-state True --unflatten-weights True --output-dtype fp16'
```

### DIFF
--- a/metaseq/scripts/reshard_fsdp.py
+++ b/metaseq/scripts/reshard_fsdp.py
@@ -18,12 +18,20 @@ logging.basicConfig(format="%(asctime)s | %(name)s | %(message)s", level=logging
 logger: logging.Logger = logging.getLogger("metaseq.scripts.reshard_fsdp")
 
 
+_STRING_TO_DTYPE: Dict[str, torch.dtype] = {
+    "fp32": torch.float32,
+    "fp16": torch.float16,
+    "bf16": torch.bfloat16,
+}
+
+
 def reshard_fsdp_checkpoints(
-    input_glob_pattern: str,
-    output_shard_name: str,
+    input: str,
+    output: str,
     num_output_shards: int = 1,
     unflatten_weights: bool = True,
     skip_optimizer_state: bool = False,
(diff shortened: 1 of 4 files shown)


metaseq/scripts/reshard_fsdp.py
```diff
@@ -135,7 +152,8 @@ def reshard_fsdp_model_weights(
             sharded_weights = []
             for weight, metadata in zip(shard_weights, shard_metadata):
                 pad = metadata["param_metadata"][idx]["params"][flat_name]["padding"]
-                sharded_weights.append(_unpad_tensor(weight[full_key], pad))
+                sharded_weight = _maybe_type(weight[full_key], dtype)
+                sharded_weights.append(_unpad_tensor(sharded_weight, pad))
```
stephenroller, line 156: Any chance the order of unpad/type matters?

stephenroller (commented):
Nice. Let's add a test for this and then lgtm.

metaseq/scripts/reshard_fsdp.py
tangbinh: I think such an order shouldn't matter here.

ruanslv (approved):

PR #561, opened 2022-12-22
Fix numpy upgrade issue

ruanslv:
Numpy 1.24.0 deprecated `np.float`, causing metaseq to throw errors: https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations

Given the number of bytes in the map I'm assuming we meant float32 here, even though np.float actually used 8 bytes instead of 4...

### DIFF
--- a/metaseq/data/indexed_dataset.py
+++ b/metaseq/data/indexed_dataset.py
@@ -105,7 +105,7 @@ _code_to_dtype = {
     3: np.int16,
     4: np.int32,
     5: np.int64,
-    6: np.float,
+    6: np.float32,
     7: np.double,
     8: np.uint16,
     9: np.uint32,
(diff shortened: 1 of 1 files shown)


ruanslv:
Test breakage is unrelated, it was already present before: https://app.circleci.com/pipelines/github/facebookresearch/metaseq/1223/workflows/06606d26-917b-422c-8717-0c316ec449a4/jobs/1714

As for pinning a version, isn't it better not to pin if we can so we can automatically get gains/advantages from future versions? Like this update, for example, it's good for our code-base as the type became clearer.

If we pin we are putting more work on ourselves later to update packages manually, which may be more painful to do if you have to do several packages in bulk?

stephenroller:
We have cython extensions that depend on numpy. I've found that accidental upgrades of numpy can break those extensions and cause very weird build errors on installation. So I would also be in favor of pinning numpy, at least to a major version like `numpy~=1.22.0`

ruanslv:
Actual crash fixed in https://github.com/facebookresearch/metaseq/pull/571. Pinned the current release and updated PR title

PR #565, opened 2022-12-23
Adding flash attention for sequence parallel

dianaml0:
**Patch Description**
Creating this PR off of #511, so it can be reviewed by @stephenroller 

The last commit (3d709dba5c4be713fd821dc4e0f6b6f90f5ead40) removes some changes from the sequence parallel code which enabled testing with world size of 1. CI is not currently running the test anyway because CI needs to be updated for the test to run. 

The forward and backward tests are passing right now. However in some cases, about .2% of the elements fail

**Testing steps**
Unit Test `gpu_tests/test_sequence_parallel_transformer_layer.py`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,6 @@ install_fairscale: &install_fairscale
             cd ~/
         fi
 
-
 install_xformers: &install_xformers
   - run:
       name: Install xFormers from Source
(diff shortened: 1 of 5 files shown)


dianaml0:
CircleCI failure not related to this PR

stephenroller:
Can we rebase for checks? Should we be concerned about the last bits of numerical differences?

dianaml0:
@stephenroller just rebased the PR, should be up to date now. The rtol and atol used are the [same ones ](https://github.com/facebookresearch/xformers/blob/e163309908ed7a76847ce46c79b238b49fd7d341/xformers/ops/fmha/common.py#L259) we use for testing in xFormers for all flash attention bwds. I do a small training run to validate, would that be useful?

dianaml0:
Looks like everything is passing now after rebasing

PR #576, opened 2022-12-31
[Cleanup] Remove MegatronTrainer

suchenzang:
Merging MegatronTrainer specific code into Trainer. 

Prepping for unifying model_parallel and non model parallel code by matching enc/dec structure.


stephenroller (approved):
noice

PR #577, opened 2022-12-31
Init for model_parallel == 1

suchenzang:


stephenroller (commented):
i tried this before. iirc the megatron code doesn't do well with it. maybe things have changed since i tried or there's something more subtle going on?

suchenzang:
@stephenroller this diff worked just fine when I tested with mp 1 🤷🏻‍♀️

stephenroller:
is it too much to ask for a test? 😅

suchenzang:
@stephenroller the 125m baseline test already caught something here, since it contains code that requires mp > 1 and transformer_lm arch to work (previously assumed that transformer_lm would never call into mp code). I've kept this test behavior as is, but I would like to redo our transformer_lm / transformer_lm_megatron branching first, because the above shouldn't even be the case in practice (another complication in our reshard logic, it seems).

Separately, we had an 8m baseline test at some point. Do you know where that went?

suchenzang:
Found it - added https://github.com/facebookresearch/metaseq/issues/578 to track this

stephenroller (approved):

PR #579, opened 2023-01-01
Initialization changes

suchenzang:
* Flag for scaling positional embedding init
* Truncate init for embeddings too
* Enable zeros init if sigma is small enough


Xirider (approved):
Looks good!

Issue #581, opened 2023-01-03
Remove IncrementalDecoder -> move logic to BaseDecoder


suchenzang:
All decodings should be incremental anyway.

PR #586, opened 2023-01-04
[Cleanup] Unify IncrementalDecoder and BaseDecoder

suchenzang:
https://github.com/facebookresearch/metaseq/issues/581


ruanslv (approved):
Did you consider keeping `IncrementalDecoder` name instead? It's more expressive, `BaseDecoder` doesn't mean much

suchenzang:
Ah.. hmm I'm thinking we'll probably get rid of the BaseDecoder abstraction too as we smush things together and simplify the codebase (the assumption here is all decodings should be incremental).  Will leave that in for now to match the BaseEncoder abstraction (currently unused lol).

Issue #590, opened 2023-01-06
md5sum mismatch in resharded OPT-175B


dpfried:
I've consolidated the 992 shards of the OPT-175B model into 8 files, following the instructions here: https://github.com/facebookresearch/metaseq/blob/main/projects/OPT/download_opt175b.md#reshard-the-shards 

but the md5sums for all of the shards don't match the ones in that README. Here are mine:

```
reshard-model_part-0.pt,ca52a4d1a3d1672b5249481169078469
reshard-model_part-1.pt,5c776fbc0f212df52fbf83b0ca4c7a64
reshard-model_part-2.pt,73e90bef8877f96674800712269149ae
reshard-model_part-3.pt,94d9442ec9fe04eab633df27d6e3ab5a
reshard-model_part-4.pt,dac742fe02fba6d500375bdd0a5113ea
reshard-model_part-5.pt,42485e9f21c759cae1f22079c0971a0b
reshard-model_part-6.pt,48ee0e983a8cde23b7544213e329afec
reshard-model_part-7.pt,ac62284b7f93d1f2a28eef6779cabeae
```

I've verified that the md5sums on all 992 individual shards that I've downloaded match the ones in https://github.com/facebookresearch/metaseq/blob/main/projects/OPT/assets/opt175b_md5sum_shards.csv .

One possible explanation is that I'm using CUDA 11.1 rather than 11.3 as specified on https://github.com/facebookresearch/metaseq/blob/main/docs/setup.md (we have a managed cluster that it's hard to install 11.3 on), but I didn't expect that to affect the model weights. Could it? I've otherwise followed the metaseq install directions exactly.

suchenzang:
@tangbinh any chance the new reshard script would change the md5sums here?

tangbinh:
> @tangbinh any chance the new reshard script would change the md5sums here?

Thank you for surfacing the issue @dpfried. @suchenzang Indeed, the new script `metaseq/scripts/reshard_fsdp.py` potentially changes the md5sums since previously we didn't reshard the optimizer state dicts as we do now. There's also an option to unflatten the weights. Let me update the docs to reflect these changes.

tangbinh:
The issue has been addressed in #591. @dpfried Please let us know if you still have concern.

guberti:
@tangbinh I don't think this issue has been resolved. When I use the latest version of the code (commit `b47f8d1`) and run the command specified in the `README`:
```bash
for j in {0..7}; do
[...]

PR #591, opened 2023-01-06
Update MD5 checkum values of FSDP consolidated shards

tangbinh:
**Summary of Changes**
As mentioned in #590, the recent changes to the FSDP resharding script (see #459) alter the MD5 checksum values when we run it with the options `--skip-optimizer-state True` and `--unflatten-weights True`. We update the doc to be consistent with those changes.

**Test Plan**
The new checksum values are computed using the following command:

```
for j in {0..7}; do md5sum /data/checkpoints/opt-175b/reshard_no_os_unflat/reshard-model_part-$j.pt; done
```

### DIFF
--- a/projects/OPT/download_opt175b.md
+++ b/projects/OPT/download_opt175b.md
@@ -43,26 +43,29 @@ done
 ```
 Note that most of our models expect to run with Model (Tensor) Parallelism. For smaller models, some
 users may find it easier to eliminate model parallelism. The checkpoints can be converted
-to eliminate use of MP with the `consolidate_fsdp_shards.py` script:
+to eliminate use of MP with the `reshard_mp.py` script:
 
 ```bash
-python metaseq.scripts.consolidate_fsdp_shards ${FOLDER_PATH}/checkpoint_last --new-arch-name transformer_lm_gpt --save-prefix ${FOLDER_PATH}/consolidated
+python -m metaseq.scripts.reshard_mp \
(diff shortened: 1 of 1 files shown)


projects/OPT/download_opt175b.md
```diff
@@ -43,26 +43,29 @@ done
 ```
 Note that most of our models expect to run with Model (Tensor) Parallelism. For smaller models, some
 users may find it easier to eliminate model parallelism. The checkpoints can be converted
-to eliminate use of MP with the `consolidate_fsdp_shards.py` script:
+to eliminate use of MP with the `reshard_mp.py` script:
 
 ```bash
-python metaseq.scripts.consolidate_fsdp_shards ${FOLDER_PATH}/checkpoint_last --new-arch-name transformer_lm_gpt --save-prefix ${FOLDER_PATH}/consolidated
+python -m metaseq.scripts.reshard_mp \
```
ruanslv, line 49: This doesn't seem to match the args from reshard_mp? https://github.com/facebookresearch/metaseq/blob/071b17aad97e7257cc026cf5d03492520bec5b91/metaseq/scripts/reshard_mp.py#L76

Am I missing something?

tangbinh: Ah, that's probably an old commit. Please see [the one](https://github.com/facebookresearch/metaseq/blob/main/metaseq/scripts/reshard_mp.py#L17) in `main` for some updates after PR #556 got merged.

ruanslv (approved):

PR #598, opened 2023-01-11
add separate config option save locally

ngoyal2707:


metaseq/trainer.py
```diff
@@ -398,13 +398,13 @@ def save_checkpoint(
                 def perform_save():
                     try:
                         logger.info(f"Beginning asynchronous torch.save to {filename}")
-                        if async_callback_fn is not None:
-                            async_callback_fn(filename)
+                        async_callback_fn(filename)
```
suchenzang, line 401: Should we gate this behind save_to_NFS logic somewhere?

metaseq/dataclass/configs.py
```diff
@@ -505,6 +505,13 @@ class CheckpointConfig(MetaseqDataclass):
     save_interval_updates: int = field(
         default=0, metadata={"help": "save a checkpoint (and validate) every N updates"}
     )
+    local_save_interval_updates: int = field(
+        default=0,
+        metadata={
+            "help": "save a checkpoint (and validate) every N updates to local SSD."
```
stephenroller, line 511:
```suggestion
            "help": "save a checkpoint (and validate) every N updates to local SSD. "
```

metaseq/trainer.py
stephenroller, line 401: why do we no longer need the None protection here? I'm going to be annoyed if we get log spam that just says Async save failed: NoneType error every 25 updates

ngoyal2707: because protection is now while submitting in the threadpool executor at line number 407.
Earlier we'd have had log spams where even with non-async saving, we'd seen these logs, it should be better now. 

The reason earlier code was written this way is becasue earlier to Zach's async saving fix, we also were doing torch.save inside the async function. Now thats outside, we dont need to submit anything to thread pool executor if async callback is None

ngoyal2707: Similar to answer on Stephen's comment, I am currently passing async_callback as None if  `save_to_NFS == False`

Xirider (approved):

PR #606, opened 2023-01-16
Add support for relative position embeddings

stephenroller:
**Patch Description**
Leaving mostly for record keeping, as experiments here were unsuccessful. Adds support for learned alibi embeddings, and in a previous version, simple offset lookups. Includes necessary changes to propagate gradients through learned positional offsets.

**Testing steps**
Ran a 7B run with `--alibi`.


metaseq/models/transformer_decoder.py
```diff
@@ -192,12 +192,39 @@ def __init__(self, args, dictionary, embed_tokens):
             if initialize_params_on_gpu:
                 alibi = alibi.cuda()
             self.alibi = torch.nn.Parameter(alibi)
```
stephenroller, line 194: I probably should've used an Embedding layer in order to get better performance.

PR #623, opened 2023-02-01
Replace optiml_paper_v1.pdf with placeholder

todpole3:
Replace optiml_paper_v1.pdf with placeholder pointing to arXiv preprint


suchenzang (approved):
@todpole3 I don't see a diff for pointing to new arXiv paper - I'm assuming you wanted to update the README too?

Approving to unblock either way!

todpole3:
@suchenzang Great catch. I changed the paper pointer in the README as well. The first change is made in the .pdf so it was not shown in the diff.

Issue #389, opened 2022-10-06
Unify model-parallel vs non model-parallel codepaths


suchenzang:
As https://github.com/facebookresearch/metaseq/issues/383 flagged, right now we have two separate codepaths for model-parallel vs non model-parallel code (aka https://github.com/facebookresearch/metaseq/blob/main/metaseq/model_parallel/models/transformer_lm.py vs https://github.com/facebookresearch/metaseq/blob/main/metaseq/models/transformer_lm.py as one example).  

This should all be unified.

stephenroller:
I looked into this and it was trickier than I expected. We have to make a fair amount of changes to megatron.

bashnick:
hey @suchenzang, I looked at the task. The model_parallel code path uses some of the models&modules from the metaseq.models path. So there are two options:
1. move everything from model_parallel path to metaseq.models path and remove model_parallel path
2. move everything to metaseq.model_parallel path and remove models&modules path.

I think, the option 1 is probably better - flattens the structure. In any case there would be quite some code refactoring. + we can probably remove all the _encoder files as well.

**Context:**
Currently we use the following models from metaseq.models in metaseq.model_parallel:
- model_parallel.models.transformer_decoder -> TransformerDecoder (from metaseq.models.transformer_decoder)
- model_parallel.models.transformer_encoder -> TransformerEncoder (from metaseq.models.transformer_encoder)
- model_parallel.models.transformer_lm -> TransformerLanguageModel (from metaseq.models.transformer_lm)
- model_parallel.models.transformer_lm -> register_model, register_model_architecture (from metaseq.models)

bashnick:
@suchenzang, I would probably do the issue #583 as a separate PR, since the [#626 ](https://github.com/facebookresearch/metaseq/pull/626) is bringing quite some changes. But I can also include it there.

suchenzang:
@bashnick yes, that makes perfect sense - the smaller you can break this up the better, but some bits might require sweeping changes to keep the diff sane, so leaving that up you!

Issue #583, opened 2023-01-03
Unify `transformer_lm_megatron` and `transformer_lm` arch


suchenzang:
In the process of unifying model-parallel codepaths (https://github.com/facebookresearch/metaseq/issues/389), we should remove the two different arch definitions.

bashnick:
**Short summary:**
Unified code paths:
- unified codepaths for model-parallel and non-model-parallel: removed the model_parallel path and merged it into models file path

Simplified inheritance layers:
- removed TransformerDecoder -> ModelParallelTransformerDecoder
- removed LanguageModel -> BaseModel
- removed TransformerDecoderLayer -> ModelParallelTransformerDecoderLayer
- removed MultiheadAttention -> ModelParallelMultiheadAttention
- removed arch transformer_lm -> transformer_lm_megatron

Updated tests accordingly:
- updated test gpu_tests/test_hf_compatibility.py to work with model_parallel

PR #626, opened 2023-02-01
Unify model-parallel vs non model-parallel codepaths

bashnick:
**Patch Description**
Unified model-parallel vs non model-parallel codepaths: merged files from metaseq.model_parallel to metaseq.models. Solving issue #389  

**Testing steps**
MP2
python -m PROJECT.sweep_baseline -g 4 -n 1 --model-size 8m --prefix test_01  --local --data /checkpoint/TEAM_NAME/datasets/consolidated/v4.0

MP1
python -m PROJECT.sweep_baseline -g 4 -n 1 --model-size 8m_mp1 --prefix test_01  --local --data /checkpoint/TEAM_NAME/datasets/consolidated/v4.0

**Results**
MP2
| Metric | Before | After | Comment |
| --- | --- | --- | --- |
| num_updates | 100 | 100 | --- |
| loss | 14.406 | 14.406 | --- |
| wps | 911142 | 909002 | <0.25% difference, noise |

MP=1
| Metric | Before | After | Comment |
| --- | --- | --- | --- |
| num_updates | 100 | 100 | --- |
| loss | 15.857 | 15.857 | --- |

**After merging**

Need to re-install metaseq:
rm -rf metaseq
copy metaseq from main branch
cd ~/rsc/metaseq
pip install --no-build-isolation -e .


suchenzang:
One more TODO (perhaps for me): 
* resume training from an existing checkpoint to see if that still works as expected
* test if inference on existing checkpoint still works as expected

(don't think either will be affected here, but will need this sanity check to be safe)

bashnick:
> 

Good point, I'll test this as well.

suchenzang:
Ok tested resuming from a previous run and things look good. From the diff, it doesn't seem like any configurations is actually impacted, so approving to move onto the next step!

suchenzang (approved):

bashnick:
> Ok tested resuming from a previous run and things look good. From the diff, it doesn't seem like any configurations is actually impacted, so approving to move onto the next step!

yes, my next diff would touch on some of the configs, when I combine the model arch: transformer_lm_megatron and transformer_lm arch. #583

PR #634, opened 2023-02-03
Unify-transformer_lm_megatron-and-transformer_lm

bashnick:
**Patch Description**
Unify transformer_lm_megatron and transformer_lm architectures. Remove *_mp.py files. Solves issue #583.

**Testing steps**
python -m pytest cpu_tests
python -m pytest gpu_tests
python -m projects.PROJECT_NAME.sweep_baseline -g 4 -n 1 --rsc --model-size 8m --tokenizer rsc --prefix $RN --local --data /checkpoint/TEAM_NAME/datasets/consolidated/v4.0


suchenzang (approved):
Niiice! Really liking how this is getting broken down across a few PRs; almost there!!

bashnick:
> Niiice! Really liking how this is getting broken down across a few PRs; almost there!!

Great, thanks! do you see that we need to further optimise the codebase here? Otherwise I planed to return to Megatron :)

suchenzang:
@bashnick Hmm so my original thinking was to get to a state where we have fewer layers of inheritance - would make it easier for researchers to trace code, since tracking methods across classes to see if they were clobbered or not.. is not fun :(  

Specifically, we currently have:
`ModelParallelTransformerLanguageModel` -> `TransformerLanguageModel` -> `LanguageModel` -> `BaseModel`. I think we can definitely remove `TransformerLanguageModel` and potentially `LanguageModel` too (warning - this will likely break some old tests, like test_sequence_generator).

Similarly, for `ModelParallelTransformerDecoder` -> `TransformerDecoder` -> `BaseDecoder`, we might be able to get rid of `TransformerDecoder`.

bashnick:
> @bashnick Hmm so my original thinking was to get to a state where we have fewer layers of inheritance - would make it easier for researchers to trace code, since tracking methods across classes to see if they were clobbered or not.. is not fun :(
Yes, totally, good point - the flatter the class structure the easier it is to debug :) I'll put it in the next PR!

PR #620, opened 2023-01-26
[WIP] code to continue training on larger number of GPUS

ngoyal2707:

PR #646, opened 2023-02-10
load-checkpoint-on-more-gpus

bashnick:
**Patch Description**
Enabling checkpoint loading on larger amount of GPUs (following logic from @ngoyal2707 PR #620). Adding test to check if loading from checkpoint created from 2 GPUs possible to start a run on 4 GPUs. Currently we support the case when new number of GPUs is multiple of the previous number of GPUs.

**Testing steps**
`cd ~/rsc/metaseq`
first run on 2 GPUs and save checkpoint:
`python3 metaseq/launcher/opt_baselines.py --prefix train.8m --model-size 8m --checkpoints-dir ./gpu_tests/test-checkpoint --tensorboard-logdir ./gpu_teststest-checkpoint --num-trials 1    --azure --num-gpus 2 --num-nodes 1 --seed 1 --local --disable-validation --max-epoch 5    --max-update 5 --benchmark`

second run with 4 GPUs and --restore-file from the previous checkpoints:
`python3 metaseq/launcher/opt_baselines.py --prefix train.8m --model-size 8m --checkpoints-dir ./gpu_tests/test-checkpoint --tensorboard-logdir ./gpu_teststest-checkpoint --num-trials 1    --azure --num-gpus 4 --num-nodes 1 --seed 1 --local --disable-validation --max-epoch 5    --max-update 5 --benchmark --restore-file ./gpu_tests/test-checkpoint/train.8m.dummy_lm.me_fp16.fsdp.zero2.relu.transformer_lm_megatron.nlay4.emb128.lrnpos.0emb_scale.tps2048.adam.b2_0.95.eps1e-08.cl1.0.lr0.001.endlr0.0001.wu50.dr0.1.atdr0.1.0emb_dr.wd0.1.ms16.uf1.mu50.s1.me5.ngpu2/checkpoint_18.pt`

2GPUs_run
4GPUs_run_from_previous

### DIFF
--- a/gpu_tests/test_checkpoint_loading_on_more_gpus.py
+++ b/gpu_tests/test_checkpoint_loading_on_more_gpus.py
@@ -1,0 +1,217 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import subprocess
+import json
+import multiprocessing
+from functools import partial, partialmethod
+import unittest
+from unittest.mock import patch
+import torch
+from metaseq.dataclass.configs import DistributedTrainingConfig
+from metaseq.launcher.opt_baselines import cli_main as sweep_cli_main
+from metaseq.cli.train import cli_main as train_cli_main
+from metaseq.launcher.opt_job_constants import Size, M
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "test requires 4 GPUs, none found")
+@unittest.skipIf(
+    DistributedTrainingConfig.distributed_world_size != 4,
+    "test requires 4 GPUs",
+)
+class TestModelParallelMP1(unittest.TestCase):
+    """
+    The test will verify that the model can started with more GPUs
+    from a checkpoint created with less GPUs. We test the first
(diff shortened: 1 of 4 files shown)


suchenzang:
Wow - good work on adding tests here! Would it be possible to also check that losses are comparable when reloading and doubling number of GPUs?

andrewPoulton:
Possibly a dumb q but I just wanted to check and I'm on mobile so it's easier to ask- does this support padding shards (if necessary)? Say (for a probably contrived example) someone wanted to go from 2->10 GPUs but the number of elts in each MP shard param was not a multiple of 5, then the original shards won't distribute evenly over the new GPUs, and would need padding.

bashnick:
> Wow - good work on adding tests here! Would it be possible to also check that losses are comparable when reloading and doubling number of GPUs?

@suchenzang, added the check for the losses are comparable after reloading on 4 GPUs.

bashnick:
> Possibly a dumb q but I just wanted to check and I'm on mobile so it's easier to ask- does this support padding shards (if necessary)? Say (for a probably contrived example) someone wanted to go from 2->10 GPUs but the number of elts in each MP shard param was not a multiple of 5, then the original shards won't distribute evenly over the new GPUs, and would need padding.

@andrewPoulton, yes, that should work. I'll test it on your example to be sure, thanks!

suchenzang (approved):

Issue #544, opened 2022-12-05
Restarts from restore-file does not work if restart occurs before a new checkpoint is written


suchenzang:
Right now, if we were to launch a job with --restore-file **and** experience a job restart shortly thereafter (before a new checkpoint is written), we fail to resume again from the --restore-file and instead start from scratch again.

This should be fixed to restart from --restore-file (though if new checkpoints are present, should still continue to restart from latest and **not** the restore file).

Xirider:
I’m working on completely reworking the load_checkpoint logic and I fixed this in my branch. But I still need to a lot of validation and testing before I’m sure that everything is working, so it will take a few weeks.

You can look at it here, but it is not completely finished yet: https://github.com/facebookresearch/metaseq/blob/peter/sort_checkpoint/metaseq/checkpoint_utils.py

PR #650, opened 2023-02-14
Rewrite of the load_checkpoint function

Xirider:
This is a rewrite of how we determine which checkpoint to load when starting/restarting a training run. (Originally there was also a refactor of how our different checkpoint paths are processed, but I seperated this out for now)

Previously we had a quite brittle logic for this, with edge cases where metaseq would not load the correct checkpoint. For example here: https://github.com/facebookresearch/metaseq/issues/544

The new logic checks first all possible sources for checkpoints (restore-file, finetune-from, local checkpoints, nfs / azure checkpoints), and assigns them priority based on their progress in training and prefers local caches.

It then takes the most recent checkpoint and copies it to local disk.

To test this you need both metaseq / metaseq-internal PR's. Here: https://github.com/fairinternal/metaseq-internal/pull/842

I tested:

multi-node and single node local start
with and without nfs cloud upload path
crashing a run and correct continuing
wiht and without finetune-from
with and without resume-file

What I didn't test yet, is if starting with an azure blob path is working.


Xirider:
Will test this on azure soon.

suchenzang:
Heads up: https://github.com/facebookresearch/metaseq/pull/646 will likely go in first since tests are passing there (after loss parity check is added).  There will probably be merge conflicts after, but hopefully not too bad.

suchenzang (approved):
Nit on getting the checkpoint saving/uploading tests to pass (may need to modify to adjust to new logic, just need to be clear why the modification to the test is needed)

Issue #632, opened 2023-02-03
Bring CM3 in!


suchenzang:
See https://arxiv.org/abs/2201.07520

PR #654, opened 2023-02-20
Implement CM3/FIM Objective

ArmenAg:
**Implements the CM3 and FIM Objective**
This was ported from the old cm3 branch which has been used for a couple of projects now. I've explicitly added support for FIM (which is a special case of CM3).

**Testing steps**
I've added a couple of tests (although I should probably add some more). I'm running a couple trainings right now to make sure both FIM and CM3 (with original Poisson parameterization) work as expected.

Fixes Issue #632


suchenzang (approved):
😻 

Nit on lint + adding copyright headers. And maybe add a reference to your paper (as a comment) for any CM3 specific logic?

PR #658, opened 2023-02-22
update Dockerfile

hisashi-ito:
**Patch Description**
There was a mismatch between [setup.md](https://github.com/facebookresearch/metaseq/blob/main/docs/setup.md) and [dockerfile](https://github.com/facebookresearch/metaseq/blob/main/Dockerfile).　so, I updated Dockefile.


suchenzang (approved):
Thanks for the patch!

PR #664, opened 2023-02-27
fix argument passing for consolidated model

lilisierrayu:
Model is not correctly initialized if it is consolidated and MP was stitched.


ArmenAg:
Lint and test failures are introduced in other places. Test failures are ok since they are specific to our project.

PR #671, opened 2023-03-09
Add scripts to convert and load Metaseq checkpoints with FasterTransformer

tangbinh:
### Summary of Changes
We add scripts to convert and load Metaseq checkpoints with FasterTransformer.

### Test Plan
See the added Markdown file.


suchenzang (approved):
Oh heck yes! If you have time - could you also include the benchmarks / stats on OPT* models using FT?

Also linking to this info from the top-level readme (https://github.com/facebookresearch/metaseq/blob/main/README.md) would be great too!

Nit: lint

ruanslv (commented):
Great work!

PR #675, opened 2023-03-12
Remove arch hack for mp1

suchenzang:


Xirider (approved):

ramakanth-pasunuru:
Would the change work with hub_utils.py (e.g., get a single model checkpoint via consolidation and load it through hub_utils)?

suchenzang:
@ramakanth-pasunuru Ah - do you know if we happen to have a test that covers hub_utils loading? If not, I think adding
```
if (
            getattr(cfg.model, "arch", None) == "transformer_lm_megatron"
            and cfg.common.model_parallel_size == 1
        ):
            cfg.model.arch = "transformer_lm_gpt"
            cfg.model._name = "transformer_lm_gpt"
```
in hub_utils will fix it (just like this logic was moved into the test_hf_compatibility test since that test was written with the above hack in place).

ramakanth-pasunuru:
I don't think we have a test with consolidated checkpoint for hub_utils https://github.com/facebookresearch/metaseq/blob/98b0b32b8f092d0240a350e2a722eccd08778a56/gpu_tests/test_hub_utils.py#L46

We can do any one of following: 
1. add the condition you mentioned to hub_utils. 
2. Maybe we can change few lines in the conversion script by setting arch, _name, and mp?
https://github.com/facebookresearch/metaseq/blob/1310b307aa7295558280d0c80976a09b592ff7d4/metaseq/distributed/stitch_fsdp_ckpt.py#L106
3. https://github.com/facebookresearch/metaseq/blob/0714529cd1342bba98c747abcccc267f943babd7/metaseq/dataclass/configs.py#L667 -- but this is not obvious/clear.

PR #668, opened 2023-03-03
working pipe

urielsinger:
mg end2end working pipe

PR #678, opened 2023-03-12
Uriel's task_ddp_backend setting

suchenzang:
See https://github.com/facebookresearch/metaseq/pull/668/

### DIFF
--- a/metaseq/trainer.py
+++ b/metaseq/trainer.py
@@ -1212,11 +1212,12 @@ class Trainer(object):
         def lower_precision(t):
             """Converts a tensor to the desired dtype based on our cfg."""
             if t.dtype is torch.float32:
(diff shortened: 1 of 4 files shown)


urielsinger (approved):
looking good! :)

metaseq/trainer.py
```diff
@@ -1212,11 +1212,12 @@ def _prepare_sample(self, sample, is_dummy=False):
         def lower_precision(t):
```
urielsinger, line 1212: From the [original PR](https://github.com/facebookresearch/metaseq/pull/668/files), need to remove line 257 of this file:
`"is_moe": getattr(self.cfg.model, "moe_freq", 0) > 0,`

suchenzang: whoops, missed that - added!

PR #679, opened 2023-03-12
Auto-delete checkpoints

suchenzang:
Adding back logic that was removed in refactoring (see: https://github.com/facebookresearch/fairseq/blob/0338cdc3094ca7d29ff4d36d64791f7b4e4b5e6e/fairseq/checkpoint_utils.py#L142-L192 and https://github.com/facebookresearch/fairseq/blob/0338cdc3094ca7d29ff4d36d64791f7b4e4b5e6e/fairseq/checkpoint_utils.py#L538-L557) for OG history.

Brought back in by @urielsinger in https://github.com/facebookresearch/metaseq/pull/668.

### DIFF
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -97,14 +97,14 @@ def save_checkpoint(
             async_callback_fn=async_callback_fn if save_to_NFS else None,
         )
 
-        if len(checkpoints) > 1:
-            # Create symlink between identical checkpoints (differing in naming for epoch/update/last).
-            for other_checkpoint in checkpoints[1:]:
-                if PathManager.islink(other_checkpoint):
-                    PathManager.rm(other_checkpoint)
-                assert PathManager.symlink(
-                    checkpoints[0], other_checkpoint
-                ), f"Failed to symlink {checkpoints[0]} to {other_checkpoint}"
+        # if len(checkpoints) > 1:
+        #     # Create symlink between identical checkpoints (differing in naming for epoch/update/last).
+        #     for other_checkpoint in checkpoints[1:]:
+        #         if PathManager.islink(other_checkpoint):
+        #             PathManager.rm(other_checkpoint)
+        #         assert PathManager.symlink(
+        #             checkpoints[0], other_checkpoint
+        #         ), f"Failed to symlink {checkpoints[0]} to {other_checkpoint}"
 
         write_timer.stop()
         logger.info(
--- a/metaseq/checkpoint_utils.py
+++ b/metaseq/checkpoint_utils.py
@@ -95,16 +95,8 @@ def save_checkpoint(
             extra_state,
             training_finished=training_finished,
             async_callback_fn=async_callback_fn if save_to_NFS else None,
+            files_to_symlink_to=checkpoints[1:] if len(checkpoints) > 1 else None,
         )
-
-        # if len(checkpoints) > 1:
-        #     # Create symlink between identical checkpoints (differing in naming for epoch/update/last).
-        #     for other_checkpoint in checkpoints[1:]:
-        #         if PathManager.islink(other_checkpoint):
-        #             PathManager.rm(other_checkpoint)
-        #         assert PathManager.symlink(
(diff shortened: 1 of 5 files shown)


metaseq/cli/train.py
```diff
@@ -521,6 +521,9 @@ def post_checkpoint_callback(cfg, num_updates, training_finished, filename):
                 f"Successfully copied {filename} to {cfg.checkpoint.cloud_upload_path}"
             )
             os.remove(filename)
+
+            # TODO[Susan]: Add symlink logic here? Check what cloud_upload_path is being used for Uriel's jobs.
```
urielsinger, line 525: we started using cloud_upload_path but there was an issue sometimes. maybe its due to the checkpoint_last issue here, but anyway we stopped using it and went back to local filesystem..

metaseq/cli/train.py
```diff
@@ -583,6 +588,16 @@ def post_checkpoint_callback(cfg, num_updates, training_finished, filename):
             except (FileNotFoundError, AssertionError) as e:
                 logger.info(f"could not upload {filename}: {e}")
 
+            # TODO[Susan]: Add symlink logic here.
+
+    # if files_to_symlink_to is not None and len(files_to_symlink_to) > 1:
```
urielsinger, line 593: still need to add logic for 'files_to_symlink_to'. what are you missing exactly from me?

metaseq/cli/train.py
```diff
@@ -492,7 +492,7 @@ def _checkpoint_add_directory(basename):
     return m[1], f"checkpoint{m[3]}"
 
 
-def post_checkpoint_callback(cfg, num_updates, training_finished, filename):
+def post_checkpoint_callback(cfg, num_updates, training_finished, filename, files_to_symlink_to):
```
urielsinger, line 495: if using `functools.partial` need to change order of `filename` and `files_to_symlink_to`

metaseq/checkpoint_utils.py
```diff
@@ -95,6 +95,7 @@ def save_checkpoint(
             extra_state,
             training_finished=training_finished,
             async_callback_fn=async_callback_fn if save_to_NFS else None,
+            files_to_symlink_to=checkpoints[1:] if len(checkpoints) > 1 else None,
```
urielsinger, line 98: can use `functools.partial` instead

metaseq/cli/train.py
urielsinger: in line 476 in this file we use `async_callback_fn` only if `cfg.checkpoint.cloud_upload_path==True`, but it should work for local storage as well, no?

metaseq/cli/train.py
suchenzang: Ah nothing needed from you - I need to put this in and see how it works on another cluster too.

metaseq/cli/train.py
suchenzang: wdym? https://github.com/facebookresearch/metaseq/blob/0c5db28049f105f60c28dda047fd195a78458239/metaseq/cli/train.py#L473-L475 both will need to be passed in

metaseq/checkpoint_utils.py
suchenzang: async_callback_fn is already a partial, so I could partial the partial to fill in the `files_to_symlink_to` arg first but that code will be very sad to follow...

suchenzang: Ya, that's what i meant by partial the partial, but seems like we don't gain much from readability either way...

metaseq/cli/train.py
suchenzang: See comment on readability in other thread (can go either way, nbd)

metaseq/checkpoint_utils.py
suchenzang: I'm more worried about interaction with async copying logic right now, since I think that's currently skipped in the Azure setup if you're only using local saves. Will test that out and update here!

davides (approved):

suchenzang:
Yay thank you for carrying the team on this @davides !!

PR #684, opened 2023-03-20
Remove mp and arch check to initialize megatron

andrewPoulton:
Remove this check - we've deprecated the non-megatron architectures, and so megatron still needs to be initialized to build mp == 1 models.


bashnick (approved):
Nice! 😍

PR #696, opened 2023-03-31
Add inference metrics

davides:
**Patch Description**
Add a facility for requesting metrics during generation with SequenceGenerator, enabled by passing `collect_metrics=True` in the request object. This will include a metrics key in the response with the following:

```
{
    "gpu_time_seconds": <total generation time>,
    "gpu_peak_mem_bytes": <allocated_bytes.all.peak reported by torch>,
    "input_tokens": <total number of tokens in the request>,
    "output_tokens": <total number of non-pad, non-eos tokens generated>,
}
```

**Testing steps**
Added a new test to `test_hub_utils`


metaseq/sequence_generator.py
```diff
@@ -618,10 +645,6 @@ def _sample_topp(self, lprobs):
         # similar issue in https://github.com/huggingface/transformers/issues/15169
         lprobs[lprobs == -math.inf] = torch.finfo(lprobs.dtype).min
 
-        # torch.multinomial gives error when every entry in lprobs are -inf
```
EricMichaelSmith, line 621: Ha - maybe this isn't a problem after all in this context?

EricMichaelSmith (approved):
Seems reasonable, thanks!

PR #698, opened 2023-04-03
Add a script to reshard FSDP-consolidated checkpoints

tangbinh:
### Summary
Since the `reshard_mp` script requires FSDP-consolidated checkpoints, we add a new script to support changing FSDP afterwards. By using`reshard_fsdp`, `reshard_mp`, and `reshard_consolidated` together, we are now able to switch from one combination of data parallel and model parallel to another.

### Test Plan
* We convert OPT 2.7B checkpoints using the following commands:
```bash
# Consolidate FSDP checkpoints (DP64, MP4 -> DP1, MP4)
for j in 0 1 2 3; do
    python -m metaseq.scripts.reshard_fsdp --input "/checkpoint/binhtang/opt-2.7b/opt-2.7b-raw/checkpoint_last-model_part-$j-shard*.pt" --output "/checkpoint/binhtang/opt-2.7b/opt-2.7b-dp1-mp4/checkpoint_last-model_part-$j-shard{i}.pt" --num-output-shards 1 --unflatten-weights True;
done

# Update MP (DP1, MP4 -> DP1, MP2)
python -m metaseq.scripts.reshard_mp --input "/checkpoint/binhtang/opt-2.7b/opt-2.7b-dp1-mp4/checkpoint_last-model_part-*-shard0.pt" --output "/checkpoint/binhtang/opt-2.7b/opt-2.7b-dp1-mp2/checkpoint_last-model_part-{i}-shard0.pt" --num-output-parts 2

# Update DP (DP1, MP2 -> DP4, MP2)
for j in 0 1; do
    python -m metaseq.scripts.reshard_consolidated --input "/checkpoint/binhtang/opt-2.7b/opt-2.7b-dp1-mp2/checkpoint_last-model_part-$j-shard0.pt" --output "/checkpoint/binhtang/opt-2.7b/opt-2.7b-dp4-mp2/checkpoint_last-model_part-$j-shard{i}.pt" --num-output-shards 4
done
```
* The converted checkpoints can be examined using the following interactive script:
```python
import logging
import random
import torch
import sys

from metaseq import options
from metaseq.dataclass.configs import MetaseqConfig
from metaseq.dataclass.utils import convert_namespace_to_omegaconf
from metaseq.distributed import utils as distributed_utils
from metaseq.hub_utils import GeneratorInterface
from metaseq.service.utils import build_logger


logger = build_logger()
logging.getLogger("metaseq.hub_utils").setLevel(logging.WARNING)


def main(cfg: MetaseqConfig):
[...]

PR #700, opened 2023-04-05
Support FSDP wrapper in SequenceGenerator

tangbinh:
### Summary
Currently, `SequenceGenerator` performs forward pass with `self.model.decoder` instead of `self.model`. This doesn't work properly with FSDP, because when wrapped with an FSDP wrapper, `self.model` is a submodule whose sharded parameters aren't all-gathered appropriately during its forward pass, resulting in errors such as the following (see Susan's comment in #698):
```
File "/shared/home/susanz/miniconda3/envs/fairseq-20220503/lib/python3.9/site-packages/torch/nn/functional.py", line 2183, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet. Caffe2 uses a lazy allocation, so you will need to call mutable_data() or raw_mutable_data() to actually allocate memory.
```
### Test Plan
* We make sure the updated `SequenceGenerator` works with and without FSDP:
```bash
python metaseq/cli/interactive.py  --merges-filename /checkpoint/binhtang/gpt2-merges.txt --vocab-filename /checkpoint/binhtang/gpt2-vocab.json --path /checkpoint/binhtang/opt-125m/opt-125m-dp4-mp2/checkpoint_last.pt --model-parallel-size 2 --distributed-world-size 8 --beam 3 --max-source-positions 4 --max-target-positions 128 --ddp-backend fully_sharded --use-sharded-sta
te
```
```bash
python metaseq/cli/interactive.py  --merges-filename /checkpoint/binhtang/gpt2-merges.txt --vocab-filename /checkpoint/binhtang/gpt2-vocab.json --path /checkpoint/binhtang/opt-125m/reshard-no-os/reshard.pt --model-parallel-size 2 --distributed-world-size 2 --beam 3 --max-source-positions 4 --max-target-positions 128 --ddp-backend pytorch_ddp
```

### DIFF
--- a/metaseq/sequence_generator.py
+++ b/metaseq/sequence_generator.py
@@ -174,7 +174,7 @@ class SequenceGenerator(nn.Module):
         # set all the forced tokens
         tokens[:, :start_step] = src_tokens.repeat_interleave(beam_size, 0)
         # compute the model predictions
-        model_out = self.model.decoder(
+        model_out = self.model(
             tokens[:, :start_step],
             incremental_state=incremental_states,
         )


suchenzang (approved):
Thank you!!

Issue #625, opened 2023-02-01
Re-release consolidated OPT / OPT-IML checkpoints


suchenzang:
After https://github.com/facebookresearch/metaseq/pull/459 and https://github.com/facebookresearch/metaseq/pull/556, we can now release updated checkpoints that are consolidated from FSDP shards with different model parallelism as well. We should update all of our checkpoints as a start to help address some of the following painpoints that users are facing:

* https://github.com/facebookresearch/metaseq/issues/595
* https://github.com/facebookresearch/metaseq/issues/594
* https://github.com/facebookresearch/metaseq/issues/574
* https://github.com/facebookresearch/metaseq/issues/567
* https://github.com/facebookresearch/metaseq/issues/475
* https://github.com/facebookresearch/metaseq/issues/407
* https://github.com/facebookresearch/metaseq/issues/233
* https://github.com/facebookresearch/metaseq/issues/211

and previous issues:
* https://github.com/facebookresearch/metaseq/issues/31
* https://github.com/facebookresearch/metaseq/issues/125

We have internal consolidated versions for 2.7B and 30B to check against, and will also need to confirm that generation looks roughly sane after consolidation.

EIFY:
I can work around it, but could the following issue be considered related as well?
* [](https://github.com/facebookresearch/metaseq/issues/552) #552

andrewPoulton:
Yeah looks like it, at least tangentially - the loading logic there could probably do with simplifying. It should be possible to identify naming convention by just reading the checkpoint directory.

PR #701, opened 2023-04-05
Update URLs to consolidated OPT checkpoints

tangbinh:
### Summary
As mentioned in #625, we consolidate and re-release OPT checkpoints to address various pain points reported by users.


suchenzang:
Yesss! Any chance we can run the interactive.py sanity check on these checkpoints to be sure consolidation happened correctly?

suchenzang (approved):
Approved to merge after sanity checks on generation passes (note that 350M model is missing layer norms so that one might be a bit off without other code changes).

PR #705, opened 2023-04-08
Porting in apex kernels

suchenzang:
Remove apex dependency, bring in relevant kernels


suchenzang:
Moving this into the megatron PR

PR #708, opened 2023-04-11
Fix target checkpoint paths for NFS symlink

davides:
**Patch Description**
Follow-up to this PR: https://github.com/facebookresearch/metaseq/pull/679

Adjust the symlink paths to match conventions when resuming with `--restore-file`

**Testing steps**
```
$ python -m tests.test_cli_train
...
.
----------------------------------------------------------------------
Ran 4 tests in 0.026s

OK
```


suchenzang:
Thanks for getting this out so quickly!!

suchenzang (approved):
Tested on existing run - looks great!

PR #715, opened 2023-04-27
Andy/drop mseq req from reshard

andrewPoulton:
The tokenization_cache in the train_iterator state_dict needs to be unpickled in a metaseq-aware env. 

Adding a flag to remove this so consolidated checkpoints load elsewhere.

Also fixes bug in named argument


suchenzang (approved):
Nit on lint!

suchenzang:
Was testing if rebasing on main could resolve gpu test failures, which apparently adds me to all commits :(  Nothing here should affect gpu tests so this is an odd failure...

PR #717, opened 2023-04-28
Generation should stop after two new lines if that is the stop criteria

Vidyaranya:
This addresses Issue 642. When the stop token is \n\n the generation should stop after generation two new lines.

Check the previous token that is generated and if it is repeated, and if the stop token contains this as the criteria, stop generating further tokens


suchenzang:
Rebase on main, tests should pass!

Vidyaranya:
> Rebase on main, and tests will probably pass!

They still seem to fail!

PR #718, opened 2023-04-28
Davides/add efficiency metrics

davides:
**Patch Description**
Optionally resource util metrics in the API response when requested.

**Testing steps**
Ran an eval that passes this flag and collects the results.


EricMichaelSmith (approved):
Seems reasonable, great to have!

ellenxtan (approved):
LGTM, thanks for adding!

PR #723, opened 2023-05-24
Add support for top-k

EricMichaelSmith:
- Add support for top-k to metaseq, to support the decoding settings we use for GPT2 and BLOOM evaluations in the ROBBIE paper

**Note:** trying LLaMa on batch size >1 gives very incoherent generations currently, but LLaMa with BS 1 gives results that are coherent and match the ROBBIE eval numbers of LLaMa run with the llm_evaluations repo

### DIFF
--- a/metaseq/hub_utils.py
+++ b/metaseq/hub_utils.py
@@ -200,6 +200,7 @@ class GeneratorInterface:
         max_tokens: List[int] = None,
         temperature: float = 1.0,
         top_p: float = -1.0,
+        top_k: int = -1,
         logprobs: int = 0,
         n: int = 1,
         best_of: Optional[int] = None,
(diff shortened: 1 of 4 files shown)


ellenxtan (approved):
Make sense, thanks for adding!

metaseq/hub_utils.py
```diff
@@ -232,6 +233,7 @@ def generate(
         max_tokens: forces EOS after this many tokens
         temperature: softmax temperature
         top_p: nucleus probability
+        top_k: top-k probability
```
davides, line 236: super-nit: maybe "top-k threshold" instead?

davides (commented):
This is great, thanks for adding!

One other place that it might be good to update is [metaseq.service.workers](https://github.com/facebookresearch/metaseq/blob/yuchen/llama_debiasing/metaseq/service/workers.py#L13). It currently tries to batch requests with similar decoding params. [This old thread](https://github.com/facebookresearch/metaseq/pull/306#discussion_r950723965) might give some context as well.

PR #727, opened 2023-06-05
Cm3 integration

urielsinger:


ArmenAg:
This has some specific logic to CM3leon project (i.e., img token conversions), are we sure we want to land this in main? Do we want to possible pick a branch and merge everything into there and periodically update to main?

metaseq/cli/train.py
```diff
@@ -144,10 +144,12 @@ def main(cfg: DictConfig) -> None:
             cfg.distributed_training,
             use_sharded_state=cfg.distributed_training.use_sharded_state,
         ):
-            model = fsdp_wrap(
-                task.build_model(cfg.model),
-                process_group=distributed_utils.get_data_parallel_group(),
-            )
+            model = task.build_model(cfg.model)
+            if not isinstance(model, FullyShardedDataParallel):
```
ArmenAg, line 148: Just to confirm, this is for loading up consolidated model for training?

metaseq/data/cm3_dataset.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
```
ArmenAg, line 4: These are changes to the cm3 objectives that i landed in scaling_racm3 correct?

ArmenAg:
Code LGTM. @suchenzang to give guidance on whether or not to land here.

metaseq/cli/train.py
urielsinger: Yes.
I added support to change the MP size during job lunch, and for that I need to wrap it in `FullyShardedDataParallel` inside the `build_model`.
As I don't want to double wrap it, I needed to add this if..

metaseq/data/cm3_dataset.py
urielsinger: Yes, exactly.

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -246,6 +281,10 @@ def _check_cm3_parameterization(self):
 
     def _create_cm3_special_tokens(self):
         self.cm3_sentinel_end = "<eoss>"
+        self.cm3_break = "<racm3:break>"
+        self.dictionary.add_symbol(self.cm3_break)
```
lilisierrayu, line 285: It's all looking great. 
We want to make a change here to recycle unused embedding index of cm3_break and sentinel for the next version. 
Should i just add a commit on top of this PR? Or do I file a different PR?

metaseq/data/cm3_dataset.py
```diff
@@ -200,24 +209,31 @@ def get_document_boundaries(self, item: torch.Tensor):
             boundaries = boundaries + [item.size(0)]
```
lilisierrayu, line 209: is get_document_boundaries() robust to the case that there is no break tokens

Issue #726, opened 2023-06-01
Possible feature and bugfix contributions from Microsoft research team's fork of Metaseq


mattmazzola:
We are a team at @microsoft Research that has a fork Metaseq repo with these additional features:

1. New pipeline task to perform Knowledge Distillation via Log Probabilities using a modified Cross Entropy implementation. 
1. Improved inference script with added functionality such as ability to output logprobs/logits.
1. Improvements to Training Stop Conditions
1. Scripts to support Teacher data generation using [Open AI Service](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/overview)
1. Documentation system using [Sphinx](https://www.sphinx-doc.org/en/master/)
    1. Documentation of Co-Teaching training process (https://arxiv.org/pdf/2305.02031.pdf)
1. Improved evaluation configuration to evaluate with different metrics depending on dataset
1. Miscellaneous Bug Fixes
    1. [`jsonl_dataset.py#_build_index`](https://github.com/facebookresearch/metaseq/blob/c16d21047d975b7a925648f38cba3190a8ef27d6/metaseq/data/jsonl_dataset.py#L135-L150) properly accounts for multi-byte characters.

## Questions

- Which of the features above would you be interested in us contributing back to Metaseq?
- Would you be able to offer assistance with the merge process?
    - For example, testing and verification of functionality for a feature PR.

We would be happy to answer any questions you have about the above components.

@tupini07

suchenzang:
@mattmazzola Sorry for delay - I've been on PTO; would be interested in all of the above contributions as they come online (deferring to you on what the best ordering here would be)!

mattmazzola:
> interested in all of the above contributions

Ok! I will talk with rest of team and see what we want to do.

We are trying to roll off our current work and transition to another project so it is not clear how much time we be able to spend these contributions. This creates a kind of trade-off / conflict between wanting the larger items for impact, but smaller items for less commitment.

> deferring to you on what the best ordering here would be

These fixes and features from our fork has some non-trivial divergence from metaseq main so it's less easy to judge how much work until we see how many merge conflicts there are. It also makes testing difficult or not possible since our infrastructure was using different dependency set running Azure Machine Learning environment.

The list above was an ordered by estimate of how impactful the PR contributions would be to Metaseq; however, given the difficulties I was trying to create PRs with inverse order to increase likelihood they merge.
Beginning with the smallest / easiest since they were least likely to break something and wouldn't rely on as much help.

[...]

PR #728, opened 2023-06-07
fix: add support for wide characters when building index of dataset files

mattmazzola:
## Context

When reading a dataset for the first time metaseq will create an "index" file that will store the byte offset in the .jsonl file where item N begins

## Issue

This calculation `cur += len(line)` was previously assuming the "length" of the read line was the offset of "next" item.
In the case that the line contains wide Unicode characters, this calculation is incorrect. `len(line)` will return 1 instead of the total bytes of the character.

### Example

An example where len gives us the wrong number with a normal string that contains wide-unicode char
The correct length is given if we use a byte-string

```bash
>>> "\u2205"
'∅'
>>> len("\u2205")
1
>>> len(b"\u2205")
6
```

## Solution

Instead of keeping a running counter this PR simplifies the implementation by querying the current position of the file reader after every line is read. This approach properly accounts for the bytes and not the characters.

Credit goes to @tupini07 for diagnosing and fixing.

One of the important bug fixes from #726 

## Testing steps

We noticed this bug in our repo when testing datasets containing these wide characters and have been using it in our pipeline for few weeks now.

### DIFF
--- a/metaseq/data/jsonl_dataset.py
+++ b/metaseq/data/jsonl_dataset.py
@@ -133,14 +133,14 @@ class JsonlDataset(torch.utils.data.Dataset):
         # and then wraps around if the epoch id goes beyond the data_subshard_count
         return (self.epoch - 1) % self.data_subshard_count
 
-    def _build_index(self, path: str):
+    def _build_index(self, file_path: str):
         """Build index of start positions of each line."""
-        logger.info(f"Building index for file: {path}")
-        f: TextIOWrapper = self._get_mmap()
+        logger.info(f"Building index for file: {file_path}")
+        file: TextIOWrapper = self._get_mmap()
 
         offsets = [0]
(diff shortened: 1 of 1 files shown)


metaseq/data/jsonl_dataset.py
```diff
@@ -132,22 +133,17 @@ def _get_subshard_id(self):
...
-        f.seek(0)
-        offsets = []
-        cur = 0
-        line_num = 0
-        while True:
-            line = f.readline()
-            if line == b"":
-                break
-            offsets.append(cur)
-            cur += len(line)
```
mattmazzola, line 148: This code is what causes the miscounting of bytes for that line of jsonl dataset and thus the index will store the incorrect byte offset.

suchenzang (approved):
Thanks for catching!

PR #729, opened 2023-06-08
fix: ensure last checkpoint is always saved, refactor training stop conditions to be computed in single location

mattmazzola:
# Issues

## 1 Inconsistent checkpoint filenames saved by trainer

In our pipeline we often have sequence of steps such as (train, reshard/unflatten, evaluate). The output files of the training become inputs to the resharding scripts. In order for the execution to work reliably the output files need to have consistent filenames, such as `checkpoint_last-model_part-0-shard0.pt`

When running metaseq.cli.train with tasks such as streaming_finetune_language_modeling there are two different stopping conditions set by --max-epochs and --max-updates. Whichever limit is hit first will cause the model stop training.

The issue is that checkpoint_last-* file is ONLY written the epoch stop condition or update stop conditions were false.
This couples the checkpoint filename with the stopping conditions

Notice `checkpoints[0]` only uses the FIRST true filename/condition

https://github.com/facebookresearch/metaseq/blob/c16d21047d975b7a925648f38cba3190a8ef27d6/metaseq/checkpoint_utils.py#L89-L99

## Goal

We want to be able to run the jobs/pipeline and change the stopping conditions without implicitly changing the output file that will be given to the subsequent commands / scripts

## 2 Training Stop was Handled in Multiple Locations

Loop condition:
https://github.com/facebookresearch/metaseq/blob/c16d21047d975b7a925648f38cba3190a8ef27d6/metaseq/cli/train.py#L209
Loop break:
https://github.com/facebookresearch/metaseq/blob/c16d21047d975b7a925648f38cba3190a8ef27d6/metaseq/cli/train.py#L212-L213

This makes it harder to reason about which condition will cause training to stop. 

# Solution

- Consolidate all training stop conditions in to `validate_and_save` and `should_stop`
    - Use `>` instead of `>=` conditions
- Change to always save `checkpoint_last*` file
    - This means there are cases it will save multiple checkpoints
        - Epoch AND last checkpoint
        - Updates AND last checkpoint
        
# Testing

I wasn't able to test since this is merging with metaseq main instead of our fork's main.
I wanted to at least share the ideas. Although changing training stop conditions can be serious, so maybe someone else can submit a small jobs to test. One with max-epochs, other with max-updates, and in both cases it saves checkpoint_last files

Related to #726

### DIFF
--- a/metaseq/cli/train.py
+++ b/metaseq/cli/train.py
@@ -452,12 +452,16 @@ def validate_and_save(
 
     if num_epoch > max_epoch:
         should_stop = True
-        logger.info(f"Stopping training due to "
-                    f"num_epoch: {num_epoch} > max_epoch: {max_epoch}")
+        logger.info(
+            f"Stopping training due to "
+            f"num_epoch: {num_epoch} > max_epoch: {max_epoch}"
+        )
     elif num_updates > max_update:
         should_stop = True
-        logger.info(f"Stopping training due to "
-                    f"num_updates: {num_updates} > max_update: {max_update}")
+        logger.info(
+            f"Stopping training due to "
+            f"num_updates: {num_updates} > max_update: {max_update}"
+        )
 
     is_epoch_save_interval = (
         end_of_epoch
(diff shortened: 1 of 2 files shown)


metaseq/cli/train.py
```diff
@@ -203,10 +203,10 @@ def main(cfg: DictConfig) -> None:
         disable_iterator_cache=True,
     )
 
-    max_epoch = cfg.optimization.max_epoch or math.inf
     train_meter = meters.StopwatchMeter()
     train_meter.start()
-    while epoch_itr.next_epoch_idx <= max_epoch:
+
+    while True:
```
mattmazzola, line 209: Usually it is practice to avoid `while True` anywhere since it relies on other code to stop the loop and it's easy to make mistakes. However, the alternative of splitting logic between loop and validate function is more complex and thus more likely for us to have issues in future.

Also, another reason the while true is not be so bad is because the original could potentially be the same condition when `max-epochs` was not provided / defined.

If not defined 
```
while epoch_itr.next_epoch_idx <= max_epoch
while epoch_itr.next_epoch_idx <= cfg.optimization.max_epoch or math.inf
while epoch_itr.next_epoch_idx <= math.inf
while true
```

PR #730, opened 2023-06-08
Fix lint / tests

suchenzang:


moyapchen (approved):

metaseq/cli/train.py
```diff
@@ -495,6 +495,8 @@ def validate_and_save(
             async_callback_fn=functools.partial(
                 post_checkpoint_callback, cfg, num_updates, should_stop
             )
+            if cfg.checkpoint.cloud_upload_path
+            else None,
```
moyapchen, line 499: Actually as a quick - what's up with this line?

moyapchen (changes requested):

metaseq/cli/train.py
suchenzang: Don't think it should've been removed in https://github.com/facebookresearch/metaseq/pull/724 , but needing to isolate the test failures coming from setuptools version now.

moyapchen (approved):

PR #731, opened 2023-06-09
feat: add .devcontainer to standardize development environment setup

mattmazzola:
⚠️ This PR likely won't work directly but I wanted to share code from our fork that may be modified to integrate ⚠️

## Issue

(This may not be 100% accurate, but was my understanding. @tupini07 Please add comments to clarify if the description below is incorrect)

Setting up the development environment for Metaseq is non-trivial. Requiring:
- Installation of Apex, Megatron, and FairScale
- Hardware that supports CUDA with drivers installed

For development and execution there is additional process of 
- building image from [Dockerfile](https://github.com/facebookresearch/metaseq/blob/main/Dockerfile)
- creating container from image 
    - mounting repo directories
    - setting arguments to utilize GPUs in containers
- [attaching VSCode to Metaseq container](https://code.visualstudio.com/docs/devcontainers/attach-container)

Making this process easier improves environment stability for current devs and reduces barrier to entry for new contributors.

## Solution

Use [VSCode .devcontainer specification](https://code.visualstudio.com/docs/devcontainers/containers) to standardize the steps above!

- Automates manual "attaching to (metaseq) container" by using metaseq image as base of .devcontainer
- Setting of non-secret environment variables
- Mounts appropriate drives
- Executes python installation

**Testing steps**

Did not test. Although these changes should not affect operation. Using .devcontainer is entirely opt-in

Related to #726


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM metaseqopt.azurecr.io/metaseq-opt-singularity:latest
```
mattmazzola, line 1: You would likely have to change the base image.
I think this image _is_ public, but there aren't guarantees about preserving it. 
Also has some other code for private cluster that you would not need and adds extra complexity

PR #733, opened 2023-06-09
feat: add inference and evaluation script with dataset transformations

mattmazzola:
⚠️ This PR is not intended to merge directly, but to share work from our fork which may be useful to Metaseq ⚠️

## Issue

1. Wanted a way to evaluate models using the same methods and commands we used for training them
1. Wanted a structured way to use different metrics or normalize differences depending on the dataset being tested
1. Wanted to use the same evaluation metrics used by [Azure Babel](https://aml-babel.com/) for comparison
1. Wanted to be able to generate few-shot prompts for inference

## Solutions

1. Add script for model inference and evaluation
    1. Add evaluation support for HuggingFace, ParlAI, COCO, Grindstone implementations of HELM metrics
    1. Update Docker file to install evaluation dependencies
        1. Unlikely to be correct modification
    1. Move _flatten_config to metaseq.utils as it used by multiple code
1. Add mappings between dataset and pipeline configuration of eval libraries, metrics, and transformation functions
    1. Metrics Example: Summarization datasets (Reddit and CNN-DM) should use evaluation libraries with ROUGE-L and BERTScore-F, but Classification datasets (HellaSwag and PIQA) should use evaluation libraries with Accuracy metric
    1. Transformation Example: HellaSwag model outputs ` (4) something something` -> `4`

1. Added necessary evaluation libraries and re-implemented some metrics
1. Add PromptGeneratror to create few-shot prompts based on configuration using [Jinja](https://jinja.palletsprojects.com/en/3.1.x/) templates

This PR is quite large so it may be hard to make sense of.
Originally was only going to be inference.py and few other modifications, but then I kept brining in missing dependencies to avoid gaps and it grew _a lot_ 🤔

## Testing

Did not test 😔

Related to: #726 

Much of this work was done by @sahajgg, @tupini07, and @anselmwang 🙏


Dockerfile
```diff
@@ -42,6 +42,16 @@ WORKDIR /build/fairscale
...
+RUN pip install \
+        aim==3.16.2 \
+        py-rouge==1.1 \
+        rouge_score==0.1.2 \
+        parlai==1.7.1 \
+        evaluate==0.4.0
+
+ENV NLTK_DATA="/usr/share/nltk_data"
+RUN python -c "import nltk; nltk.download('punkt', download_dir='${NLTK_DATA}')"
+
```
mattmazzola: This likely isn't the correct place to make this change.

It is only snippet from our whole [Dockerfile](https://github.com/facebookresearch/metaseq/pull/731/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) which adds the evaluation libraries

metaseq/data/datasets/dataset_configurations.py
```diff
@@ -0,0 +1,84 @@
+from metaseq.data.datasets import e2e_transformers, hellaswag_transformers, piqa_transformers, reddit_transformers, cnn_dm_transformers
+from metaseq.data.datasets.types import CommonDatasetConfiguration, DatasetConfiguration, DatasetConfigurationTeacherGenerated, DatasetModelConfig, DatasetModelHooks, DatasetTeacherGeneratedDataHooks, IdentityDict
+
+# Visual diagram of where hooks/functions are called during inference or data generation
+# https://excalidraw.com/#json=zoAk_TdynBHQnP9vZufGm,ekcVg_HqiF79cAp58_HKRQ
```
mattmazzola, line 5: This visualization may be important for understanding

Dockerfile
```diff
@@ -42,6 +42,16 @@ WORKDIR /build/fairscale
 RUN git checkout fixing_memory_issues_with_keeping_overlap_may24
 RUN pip3 install -e .
 
+RUN pip install \
+        aim==3.16.2 \
```
tupini07: I think we can remove `aim` since it's more for monitoring of jobs than evaluation :)

metaseq/generation_metrics/__init__.py
```diff
@@ -0,0 +1,19 @@
+import importlib
+import os
+
+from .coco_metrics import CocoMetrics
+from .hf_metrics import HFEvaluateMetrics
+from .metrics import GenerationMetrics
+from .parlai_metrics import ParlAiMetrics
+
+__all__ = ["GenerationMetrics", "HFEvaluateMetrics", "ParlAiMetrics", "CocoMetrics"]
```
tupini07, line 9: I think we can remove this and everything below. We're not including the `tasks` directory in this PR, and it's also no longer used.

metaseq/data/prompt_generator.py
```diff
@@ -0,0 +1,314 @@
...
+        self,
+        prompt_template_s: Optional[str] = None,
+        n_fewshot_samples: int = 0,
+        few_shot_data_glob_path: Optional[str] = None,
+        fewshot_sample_method='random',
+        rng_seed=42,
+        max_tokens: int = None,
+        shots_to_input_truncation_percentage: float = 0.5,
+        tokenizer_vocab_file_path="/mnt/input_data_dir/pretrained_models/OPT/dependencies/gpt2-vocab.json",
+        tokenizer_merges_file_path="/mnt/input_data_dir/pretrained_models/OPT/dependencies/gpt2-merges.txt",
```
tupini07, line 45: If `Metaseq` has a standardized path for the _vocab_ and _merges_ files then we'll need to replace them here :) If not we might need to remove the default value.

metaseq/generation_metrics/grindstone_metrics.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def evaluate_closure(pred: str, references: List[str]) -> Dict[str, float]:
+        # https://github.dev/stanford-crfm/helm/blob/a1c5e4293d6b475aca567abaadf4eae8c978bd4e/src/helm/benchmark/metrics/summarization_metrics.py#L150-L152
+        precision, recall, f1score = scorer.score(cands=[pred], refs=[references])
+        return {"BERTScore-P": precision[0].item(), "BERTScore-R": recall[0].item(), "BERTScore-F": f1score[0].item()}
+
+    return evaluate_closure
+
+
+def _buld_accuracy_scorer() -> EvalFunctionType:
```
tupini07: typo

```suggestion
def _build_accuracy_scorer() -> EvalFunctionType:
```

metaseq/generation_metrics/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+
+# automatically import any Python files/directories in the tasks/ directory
+tasks_dir = os.path.join(os.path.dirname(__file__), "tasks")
+
+if os.path.exists(tasks_dir):
+    for item in os.listdir(tasks_dir):
+        path = os.path.join(tasks_dir, item)
+        if (not item.startswith("_") and not item.startswith(".") and (item.endswith(".py") or os.path.isdir(path))):
+            task_name = item[:item.find(".py")] if item.endswith(".py") else item
+            module = importlib.import_module("metaseq.generation_metrics.tasks." + task_name)
```
tupini07: I think we can remove this. 

The `__all__` is not used because we're never importing `*` from `generation_metrics`. We're also not including the `tasks` directory in this PR, and it's also no longer used so there's no need to :)

```suggestion
```

metaseq/utils.py
```diff
@@ -93,6 +124,10 @@ def make_positions(tensor, padding_idx: int):
     return (torch.cumsum(mask, dim=1).type_as(mask) * mask).long() + padding_idx
 
 
+def strip_pad(tensor, pad):
```
tupini07: thought: this could be renamed to `strip_token`, since there's no real constraing that `pad==<PAD>`

metaseq/cli/inference.py
```diff
@@ -0,0 +1,764 @@
...
+                    f.write(json.dumps(result, ensure_ascii=False) + "\n")
+
+    except Exception as e:
+        import traceback
+        logger.error(f"Error in writing inference results to file.")
+        logger.error(f"stack trace: {traceback.format_exc()}")
+
+
+def generate(cfg: MetaseqConfig, args: argparse.Namespace):
+    global TOTAL_TOKENS_GENERATED
```
tupini07, line 294: now that I think about it, we should probably set this to `0` here, in case we call `generate()` more than once during the lifetime of the process.

metaseq/cli/inference.py
```diff
@@ -0,0 +1,764 @@
...
+    """
+    source_key = "text" if "text" in json else "src"
+    assert source_key in json, "json must contain a 'text' or 'src' field"
+
+    target_key = "tgt" if "tgt" in json else None
+
+    # NOTE: if you pass in a delimiter, we assume the target is already included
+    # in the `text/src` field, and we split on the delimiter to get the prompt
+    # and target respectively. That means that if you pass in a delimiter, you
+    # should not pass in a `tgt` field since it will be ignored.
```
tupini07: It will actually raise an `AssertionError`

```suggestion
    # should not pass in a `tgt` field.
```

or 

```suggestion
    # should not pass in a `tgt` field since it would be ignored.
```

metaseq/cli/inference.py
```diff
@@ -0,0 +1,764 @@
...
+                max_seq_len=MAX_SEQ_LEN,
+                max_prompt_len=args.max_prompt_len,
+                truncation=args.truncation,
+                prompt_end_text=args.prompt_end_text,
+            ),
+            epoch=1,
+            data_subshard_count=1,
+            output_raw_items=args.copy_input_to_output
+        )
+        for data in json_dataset:
```
tupini07, line 351: In the spirit of transparency, just wanted to clarify that here we are actually loading the whole dataset into RAM, which means this won't work if you try to inference on a dataset that doesn't fit in memory.

We had it in our backlog to load this as a stream, but never got around to it.

metaseq/cli/inference.py
```diff
@@ -0,0 +1,764 @@
...
+    distributed_utils.global_barrier()
+
+    if distributed_utils.get_global_rank() == 0:
+        progress_bar.close()
+
+        if args.merge_preds_on_all_ranks:
+            unified_output_prediction_file = os.path.join(str(cfg.common_eval.results_path), "all_prediction_results.jsonl")
+            # TODO this means that files are not sorted! In previous
+            # implementation the predictions were sorted by `instance_idx`.
+            # Confirm if this is really necessary (it isn't as far as I've seen)
```
tupini07: I we can remove this comment since it's no longer relevant. If I'm not mistaken, it's a leftover from one of my PRs and I never intended to merge it 😅

```suggestion
```

tupini07 (commented):
left some comments :)

Dockerfile
mattmazzola: Ok, yea they also install this in setup.py `aim>=3.9.4` so I think it's redundant and would distract from PR conents

metaseq/generation_metrics/__init__.py
mattmazzola: Ok, removed

metaseq/generation_metrics/grindstone_metrics.py
mattmazzola: typo fixed

metaseq/utils.py
mattmazzola: Ok, renamed to strip_token and renamed param pad to token

I don't think the function is actually used in this PR

metaseq/cli/inference.py
mattmazzola: Ok, added another assignment to 0

metaseq/cli/inference.py
mattmazzola: Hmm, I chose the top option of removal

metaseq/cli/inference.py
mattmazzola: Ok, added TODO about streaming and comment to clarity

metaseq/cli/inference.py
mattmazzola: Ok, removed

PR #736, opened 2023-06-12
feat: add soft distillation

mattmazzola:
⚠️ This PR is not intended to be merged directly. Purpose to share features that may be useful for Metaseq ⚠️

# Background

One of the main goals for our project's fork was to implement "soft" distillation (training on set of logprobs rather than correctness of token class) and to measure the efficacy of this technique compared to normal finetuning

From our docs:

> The motivation for training on log probabilities rather than token classes is to pass as much knowledge from the teacher to the student as possible. [... By the teacher providing] log probabilities of other tokens in the vocabulary [we expect] the student better learn to represent the teacher’s knowledge.

# Issue

- Soft Distillation was not implemented

# Solution

- Add new pipeline task `streaming_distillation_language_modeling`
    - Add new criterion `vocab_parallel_soft_cross_entropy` (Note: **Soft**)
        - Considers multiple possible predictions for each token of the target sequence
    - Adds new parameters 
        `--task streaming_distillation_language_modeling`
        `--distillation-mode logprobs_distillation`
       `--criterion vocab_parallel_soft_cross_entropy`

# Testing

Did not test 

Related to #726 

This feature was implemented by @anselmwang and @clarissesimoes


metaseq/criterions/vocab_parallel_soft_cross_entropy.py
```diff
@@ -0,0 +1,163 @@
...
+            sum_exp_logits, op=torch.distributed.ReduceOp.SUM, group=get_tensor_model_parallel_group()
+        )
+
+        # Loss = log(sum(exp(logits))) - predicted-logit.
+        target_weights = target_predictions.exp()
+        loss = ((torch.log(sum_exp_logits).unsqueeze(dim=-1) - predicted_logits) * target_weights).sum(-1)
+
+        # Store softmax, top_logprobs-mask and masked-top_logprobs for backward pass.
+        softmax = exp_logits.div(sum_exp_logits.unsqueeze(dim=-1))
+        ctx.save_for_backward(softmax, target_mask, masked_target, target_weights)
```
mattmazzola, line 66: If I remember, the majority of work was in this section of code. @clarissesimoes can you confirm? 
If not, can you make a comment to call out other notable places of code they should pay closer attention to

metaseq/criterions/vocab_parallel_mse_loss.py
```diff
@@ -0,0 +1,156 @@
...
+    from megatron.mpu.initialize import get_tensor_model_parallel_group
+    from megatron.mpu.initialize import get_tensor_model_parallel_rank
+    from megatron.mpu.initialize import get_tensor_model_parallel_world_size
+    from megatron.mpu.utils import VocabUtility
+    has_megatron_submodule = True
+except (ImportError, ModuleNotFoundError):
+    has_megatron_submodule = False
+
+
+class _VocabParallelMSELoss(torch.autograd.Function):
```
mattmazzola, line 18: From what I remember, we implemented this thinking it would work for Logits AND Logprobs, but it only works for Logits. Then because we could only get logprobs from OpenAI model output and couldn't convert to logits this loss effectively became unused.

However, we left it in because the implementation may be valuable for other applications where logit values are available

metaseq/criterions/vocab_parallel_cross_entropy.py
```diff
@@ -25,15 +26,24 @@ def forward(self, model, sample, reduce=True):
...
+        # Compute token accuracy across vocab parallel groups
+        ignored_token_ids = [self.padding_idx, self.eos_token_id] if has_mask else []
+        token_accuracy = utils.vocab_parallel_token_accuracy(
+            vocab_parallel_logits=logits, target=target_token_ids, ignored_token_ids=ignored_token_ids
+        )
+
+        # When using target loss only, use num tokens in target as the sample_size
         # See StreamingSrcTgtDataset
         sample_size = (
             sample["ntokens_target"]
```
clarissesimoes, line 49: This should be 
```python
sample_size = (sample["ntokens_target"] if "ntokens_target" in sample else sample["ntokens"])
```

metaseq/models/transformer_decoder.py
```diff
@@ -59,10 +60,10 @@ def log_weight_stats(tensor, name):
     )
 
 
-class ModelParallelTransformerDecoder(BaseDecoder):
```
clarissesimoes, line 62: I don't recall changing this class in our implementation, but this might have been changed by Yu or Sahaj. It would be a good idea to include Sahaj in this review too.

metaseq/models/transformer_decoder.py
```diff
@@ -491,47 +511,54 @@ def buffered_future_mask(self, tensor, input_tokens=None):
...
+            is_sequence_parallel,  # sequence_parallel
         )
-        if not features_only:
-            x = self.output_layer(x)
+        # Gather output if model is in inference mode (i.e. evallm or generation) cause both are not yet compatible with
+        # parallel vocab embeddings
+        criterion = getattr(self.args, "criterion")
+        is_parallel_criterion = criterion.find("vocab_parallel") != -1
+        if not is_parallel_criterion or getattr(self, "inference", False):
+            x = gather_from_tensor_model_parallel_region(x).contiguous()
```
clarissesimoes, line 548: This is the only change that I confirm that I've made in this class.  Please confirm if the others should be double checked with Sahaj and Yu. My concern is that we might be reverting changes important in the original code in the rest of the diff

tests/test_streaming_distillation_language_modelling_task.py
```diff
@@ -0,0 +1,117 @@
...
+from metaseq.tasks.streaming_distillation_language_modeling import StreamingDistillationLanguageModelingTask
+
+
+class TestResamplingDataset(unittest.TestCase):
+
+    def setUp(self):
+        self.dict_args = {}
+        self.dict_args['hf_tokenizer'] = ""
+        self.dict_args['vocab_filename'] = "./tests/tokenizer_files/gpt2-vocab.json"
+        self.dict_args['merges_filename'] = "./tests/tokenizer_files/gpt2-merges.txt"
```
clarissesimoes, line 15: Please add these two files. They're required to enable tests here

clarissesimoes (changes requested):
I think one the most valuable things that we've added is the documentation, as both parallel vocabulary cross entropy and soft cross entropy can be hard to understand due to the fact of both of them applying simplifications in the formulas, all of them explained in the docs. Is that a way to bring the MD files that we've created to this diff?

metaseq/models/transformer_decoder.py
mattmazzola: I don't think we can modify reviewers since we are not contributors/admins on repo.
However, I think we can mention them here @anselmwang @sahajgg

tests/test_streaming_distillation_language_modelling_task.py
mattmazzola: Ok, I added the files

metaseq/criterions/vocab_parallel_cross_entropy.py
mattmazzola: I think Github was collapsing part of the code which made it appear as 

```
sample_size = (
   sample["ntokens_target"]
)
```
but expanded it is equivalent to your line
```
sample_size = (
   sample["ntokens_target"]
   if "ntokens_target" in sample
   else sample["ntokens"]
)
```

Meta uses a different formatter and I was trying to reduce the changes in the reviews by using their formatting when they are the same

metaseq/criterions/vocab_parallel_soft_cross_entropy.py
mattmazzola: Also, there is a [documentation file](https://github.com/facebookresearch/metaseq/pull/736/files#diff-100e01eb637c0d62f554384f884a21acda6ce6b85bf589e4155e83fd2211f504) included in PR to help explain this code.

metaseq/criterions/vocab_parallel_mse_loss.py
clarissesimoes: We've tested MSE Loss with logprobs but training never converged. MSE Loss can be used if the input is teacher logits, though

metaseq/criterions/vocab_parallel_soft_cross_entropy.py
clarissesimoes: I confirm, and I'd also mention the file [metaseq/tasks/streaming_distillation_language_modeling.py](https://github.com/facebookresearch/metaseq/pull/736/files#diff-2e72edf07bf0c27b64018b7cfebc34455a23e53cdb09fff9cba6785f440d484f) as equally important, as data preprocessing and masks are slightly different for distillation when compared to finetuning

anselmwang (approved):

PR #739, opened 2023-06-27
[fix] Set upper limit for aim version

alberttorosyan:
With the upcoming major release of Aim (https://github.com/aimhubio/aim) Aim SDK will not be backwards compatible.
To prevent any failures on metaseq side, setting upper limit for the install Aim version


alberttorosyan:
We are going to release the major release of Aim soon. Would be great if this PR can be merged beforehand, to prevent any failures.

PR #746, opened 2023-07-14
Implement JSONLDataset level partitioning

ArmenAg:
# High Level Change
We augment where partitioning of the data happens to be earlier in the process saving a lot of compute when it comes to running tokenization. This should allow us to scale WPS past the ceiling (~3.5M) we were seeing earlier.

# How Did We Implement?
We utilize the virtual sub-sharding implemented for JSONLDataset instead as a partitioning function (this means we cannot do sub-sharding, but no one is really using this anyway, assertions have been added). We had to adapt some of the load_dataset function signatures but overall this is a very easy change.

# What We Do Now
The current way we generate documents is the following. Each GPU does the following:
- Enumerate all documents
- Tokenize all documents
- Block by sequence length
- Each GPU selects the block that corresponds to it’s ID.

Untitled Diagram-Page-1 drawio

Pros:
- Blocks are completely randomized on all GPU’s
- Fully reproducible. Can restart a run and follow the same exact selection process.

Cons: 
- Every GPU has to tokenize EVERY sample, which might be the cause the lack of GPU saturation.

In the current change, each GPU does the following:
1. Enumerate all documents
2. Select documents that correspond to it’s ID
3. Tokenize current GPU documents
4. Block by sequence length
5. Each GPU simply enumerates blocks

Untitled Diagram-Page-2 drawio (1)

Pros:
- Fully reproducible. Can restart a run and follow the same exact selection process. But ONLY when world size stays the same throughout training.
- Tokenization only occurs on the documents you actually use on the current GPU

Cons:
- Are there any? 

# Questions
Q. What happens if one GPU runs out of documents before others?
A. Same thing that happens in the previous code when one GPU is out of blocks, we stop training on current shard. The amount of data we'll be leaving behind is miniscule.

Q. Well if we're leaving data behind is this not problematic for consistent validation?
A. Yes for validation we don't _really_ care about WPS so we're okay using the old implementation that will use all of the data (except maybe for the last batch).

### DIFF
--- a/metaseq/tasks/streaming_language_modeling.py
+++ b/metaseq/tasks/streaming_language_modeling.py
@@ -8,33 +8,33 @@ on-the-fly tokenization.
 """
 
 import logging
-import random
 import os
+import random
 import re
+import zipfile
+from collections import defaultdict, namedtuple
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional
-from collections import defaultdict
-import zipfile
-from metaseq.utils import print_r0
+
 import numpy as np
 import torch
-from omegaconf import II
 
 from metaseq.data import (
+    data_utils,
     Dictionary,
+    iterators,
     JsonlDataset,
     PartitionedStreamingDataset,
     ResamplingDataset,
     StreamingSrcTgtDataset,
-    data_utils,
-    iterators,
 )
-
-from metaseq.dataclass import MetaseqDataclass
-from metaseq.tasks import LegacyTask, register_task
-from metaseq.data.document_to_sequence import DocumentToSequenceDataset
 from metaseq.data.cm3_dataset import CausalMaskedDocumentToSequenceDataset
-from metaseq.dataclass import ChoiceEnum
+from metaseq.data.document_to_sequence import DocumentToSequenceDataset
+
+from metaseq.dataclass import ChoiceEnum, MetaseqDataclass
+from metaseq.tasks import LegacyTask, register_task
+from metaseq.utils import print_r0
+from omegaconf import II
 
 try:
     from tokenizers import ByteLevelBPETokenizer, Tokenizer
@@ -50,6 +50,10 @@ DEFAULT_MULTICORPUS_MAX = -1
 
 LANGUAGE_MODELING_MODE = ChoiceEnum(["standard", "cm3", "racm3"])
 CM3_MODE = ChoiceEnum(["poisson", "fixed", "fim"])
+
+DatasetWithShardInformation = namedtuple(
+    "DatasetWithShardInformation", ["dataset", "is_sharded", "shard_id", "num_shards"]
+)
 
 
 def map_old_image_token_to_new_image_token(text):
@@ -263,6 +267,10 @@ class StreamingLanguageModelingTask(LegacyTask):
         else:
             self.dictionary.pad_to_multiple_(8)
 
+        assert (
+            self.args.data_subshard_count == 1
+        ), "We utilize subsharding as a way to do faster data processing therefore do not allow for another layer of subsharding."
+
     def _check_cm3_parameterization(self):
         assert (
             self.args.cm3_lambda_sentinel_tokens > 0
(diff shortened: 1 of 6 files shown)


metaseq/tasks/streaming_language_modeling.py
```diff
@@ -605,32 +620,42 @@ def load_dataset(self, split: str, epoch=1, combine=False, **kwargs):
...
+                    dataset,
+                    block_size=self.args.tokens_per_sample + 1,
+                    break_mode=self.args.sample_break_mode,
+                    drop_last=(split == "train"),
+                    padding_idx=self.source_dictionary.pad(),
+                    seed=self.args.seed,
+                ),
+                is_sharded=is_sharded,
+                shard_id=shard_id,
+                num_shards=num_shards,
```
sriniiyer: Move to outside if-else to remove code duplication?

self.datasets[split] = DatasetWithShardInformation(self.datasets[split] , is_sharded ....)

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -511,7 +519,14 @@ def get_previous_shard_str(self, epoch, split):
         prev_shard_str = shards[shard_idx - 1] if shard_idx > 0 else None
         return prev_shard_str
 
-    def load_dataset(self, split: str, epoch=1, combine=False, **kwargs):
+    def load_dataset(
+        self, split: str, epoch=1, combine=False, num_shards=-1, shard_id=-1, **kwargs
+    ):
+        is_sharded = True
+        if num_shards < 0 or shard_id < 0:
+            num_shards = 1
```
arbabu123, line 561: assert num_shards < 0 && shard_id < 0  ? :P

metaseq/tasks/streaming_language_modeling.py
```diff
@@ -51,6 +51,10 @@
 LANGUAGE_MODELING_MODE = ChoiceEnum(["standard", "cm3", "racm3"])
 CM3_MODE = ChoiceEnum(["poisson", "fixed", "fim"])
 
+DatasetWithShardInformation = namedtuple(
+    "DatasetWithShardInformation", ["dataset", "is_sharded", "shard_id", "num_shards"]
```
arbabu123, line 55: where exactly are we using this extra fields ? for eg - num_shards : in places like get_batch_iterator its passed as parameter?

arbabu123: i saw dataset.is_sharded being used

arbabu123: is it possible to add some flag to jsonl dataset itself and remove this named tuple completely?

ArmenAg: We pass JsonlDataset into ConcatDataset. I would rather have a wrapper over a dataset rather than have to propagate this information through all datasets. WDYT?

metaseq/tasks/streaming_language_modeling.py
ArmenAg: sg

metaseq/tasks/streaming_language_modeling.py
arbabu123: so is ("shard_id", "num_shards") used somewhere else?

metaseq/tasks/streaming_language_modeling.py
ArmenAg: some parts of the code pass in None now (for example validation).

metaseq/tasks/streaming_language_modeling.py
ArmenAg: added extra assertions using this to insure consistency between iterator/dataset

ArmenAg:
To run test: `pytest cpu_tests/test_streaming_language_modeling_task.py -s -k test_load_partitioned_dataset`

sriniiyer (approved):
- Tested that wps is stable
- Tested that no significant change in ppl before and after this change

lgtm!

PR #747, opened 2023-07-15
adding weighted

lilisierrayu:
adding support that we can decide down/up sampling of each dataset with a dictionary, for example. 
    {
    "arxiv":1.14,
    "b3g":3.15,
    "c4":6,
    "ccnet_new":28.3,
    "edouard_cc_20220927_new":24.7,
    "github_oss_with_stack":4,
    "stackexchange":0.88,
    "wikipedia":3.5,
    }


ArmenAg:
Could you link to this PR the xlformers implementation just to make sure we're 1:1?

PR #749, opened 2023-07-21
allow hubutils to support generators argument

ramakanth-pasunuru:
**Patch Description**
Allow hubutils to support user specific generator as an extra argument. 

**Testing steps**
Ran a simple eval and it worked fine.


lilisierrayu (commented):
LGTM

lilisierrayu (approved):

